PR #201, opened 2015-05-22
Topic/new locus search

nmenda:
merge new locus search form using jQuery data tables with master


nmenda:
changed search to start with the locus table and display string of allele symbols. 
If the test work , it needs to be merged with master

PR #269, opened 2015-10-28
accessions and controls on trial detail page have links to their stock detail page

nickmorales:
trial_detail_accession_link


bellerbrock:
nice solution!

lukasmueller:
Would be great to add a selenium2 test for the stock detail page.

Issue #457, opened 2016-04-20
compare and visualize training and selection population breeding values


isaak:
suggested by Ben Hayes.

PR #577, opened 2016-07-07
Topic/solgs/compare gebvs training selection

isaak:
--visualizes selection gain by creating visual comparison of training and selection populations' GEBVs.
--calculates z-scores for observations and corresponding p-values.
--draws normal curves and marks means of each population.


isaak:
closes #457.

PR #794, opened 2016-12-16
Topic/improve pheno download speed

bellerbrock:
@nickmorales This is the implementation of a fast pheno download using the materialized views like we discussed, in parallel with the original complete query. It's implemented so the user can chose which type, and I've added a test.

Do you mind taking a look and seeing if anything should be changed before I merge? In particular I'm not sure which query type the BrAPI calls should use by default.


nickmorales:
Looks OK to me but I don't see a test. You could add tests to the phenotypes.t. at the bottom of that file there are 3 tests. You could run the same tests against the fast method and check if they're the same

bellerbrock:
Ah, my bad, I forgot to add the new test file to the last commit. Adding it now, and I'll look to see its tests can be moved to phenotypes.t

nickmorales:
Maybe a test where you first add new phenotypes, then refresh the view, then do a phenotype search using the native and view methods. Then the two search methods should return the same result.

The phenotypes.t test already adds a bunch of phenotypes and tests the native search, so you'd only need to append to the end of that test a refresh of the matviews and then a check on the view search results.

PR #862, opened 2017-02-14
during phenotype upload there is an option available to curators to b…

nickmorales:
during phenotype upload there is an option available to curators to bypass the variable_of constraint for traits


nickmorales:
this does not seem to be the best long term solution

PR #884, opened 2017-02-27
Topic/phenotyping barcode

aco46:
@nickmorales : could you please take a look at the plot phenotyping barcode implementation.


lib/SGN/Controller/Barcode.pm
```diff
@@ -117,7 +117,7 @@ sub barcode_qrcode_jpg : Path('/barcode/tempfile') Args(2){
```
nickmorales: See comment in SGN::Controller::QRcode

aco46: fixed

PR #897, opened 2017-03-03
Topic/accession usage

titima15:


mason/accession_usage.mas
```diff
@@ -0,0 +1,90 @@
+<%args>
```
nickmorales: move this into a directory like /mason/stock/accession_usage.mas

mason/accession_usage.mas
```diff
@@ -0,0 +1,90 @@
...
+  jQuery('#accession_usage_trials').DataTable({
+
+     'ordering' : false,
+     'ajax':'ajax/accession_usage_trials'
+  });
+
+});
+
+
+jQuery(document).ready(function(){
```
nickmorales: Its cleaner to put these all in a single document ready

PR #899, opened 2017-03-06
Generic units ontology

nickmorales:
a generic unit ontology that can certainly be refined and expanded.

relates to issue #860


nmenda:
looks like a good start!

PR #965, opened 2017-04-13
Topic/fix multi-location trial error

aco46:
@lukasmueller @nickmorales @bellerbrock 
Please, could someone assist in reviewing this branch. It's only a couple of lines that adds trial_type to multi-location trial.

I need to use it in yam test for workshop this evening.


aco46:
-also merged master in the most recent commit.

PR #968, opened 2017-04-14
Topic/blast_slurm -> new feaures for BLAST graph

noefp:


nickmorales:
Please use another branch like blast_changes...
The changes to run blast with slurm disappeared from this branch, but would be nice to have them ready for when we do move to slurm...

PR #978, opened 2017-04-20
check trait name before querying for id..

isaak:
fixes list type prediction bug


isaak:
misdiagnosed ...not a bug...

Issue #817, opened 2017-01-12
test different possible materialized view updates to accomodate post-composed phenotypes


bellerbrock:
From Tuesday's discussion - the plan is, using topic/postcomposing branches, to load demo post-composed data in the fixture. The demo data doesn't need to be complicated, terms to compose can come from the cassava trait ontology and by loading simple time and unit ontologies.

Then, using the demo data, test out the possible solutions to storing and accessing post-composed phenotypes in the materialized views. These include storing them with a single group_id, storing the multiple cvterm_ids within a json string, or storing individual rows for each separate term.

bellerbrock:
After looking at some demo data today, we've come up with a plan going forward for the materialized views and the wizard. The plan is, instead of jsonb arrays, to a add a trait_group view, and a group_id column to the phenoview. And a new trait_components category in the wizard and binary views.

Issue #889, opened 2017-03-01
Add support for composing existing traits with time terms


bellerbrock:
Right now trait composition in topic/postcomposing_alt only works with a minimum set of terms from the trait_component ontologies. We should also let users combine exisitng traits with time terms.

Issue #888, opened 2017-03-01
create generic unit and time ontologies for all dbs


bellerbrock:
These can just be generic versions of the cassbase unit and time ontologies, with an extra months category added to the time ontology.

PR #987, opened 2017-04-21
Topic/postcomposing alt

bellerbrock:
implements postcomposing and updates materialized views. 
depends on patches in db/00073 and db/00074.
Ontologies for composing must be assigned cvprops using bin/load_composed_cvprops.pl


bellerbrock:
closes #817 #888 #889 #890

PR #992, opened 2017-04-24
Topic/solgs/job submission fix

isaak:
--fixes job submission and emailing bug

### DIFF
--- a/js/solGS/solGS.js
+++ b/js/solGS/solGS.js
@@ -519,7 +519,7 @@ solGS.waitPage = function (page, args) {
 	    success : function(response) {
 		if (response.result) {
 		    runAnalysis(profile);
-		    confirmRequest();
+		    ////confirmRequest();
 		    
 		} else { 
 		    jQuery('<div />', {id: 'error-message'})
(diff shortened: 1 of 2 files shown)


js/solGS/solGS.js
```diff
@@ -566,6 +566,42 @@ solGS.waitPage = function (page, args) {
 	    type    : 'POST',
  	    data    : profile,
 	    url     : '/solgs/run/saved/analysis/',
+	    success : function(response) {
+		if (response.result.match(/Submitted/)) {
+		    confirmRequest();
+		} else {
+		    jQuery('<div />')
+			.html('Error occured submitting the job. Please contact the developers.' + "\n\nHint: " + response.result)
+			.dialog({
```
nickmorales: maybe use bootstrap modals instead to standardize how dialogs look in the website.

isaak: Good idea..do we have standardized style somewhere that I can follow up?

nickmorales: https://www.w3schools.com/bootstrap/bootstrap_modal.asp
The modal is defined in html in a mas file, then you can use jQuery or a html data trigger to open the modal. Then you can use jQuery to run events when buttons are clicked (e.g. the OK button)

PR #1067, opened 2017-06-12
Topic/trait variable search

nmenda:
search for cvterms that have variable_of relationship with a parent trait name
The trait_cv_name has to be defined in the conf file


lib/CXGN/Trait/Search.pm
```diff
@@ -0,0 +1,193 @@
...
+    }
+    my $trait_name_is_exact = $self->trait_name_is_exact;
+    my $sort_by = $self->sort_by;
+    my $order_by = $self->order_by;
+
+    my $trait_rs;
+    
+    if ($is_variable) { 
+	# pre-fetch some information; more efficient
+	my $variable_of_cvterm_id = SGN::Model::Cvterm->get_cvterm_row($schema, 'VARIABLE_OF', "cassava_trait")->cvterm_id();
```
nickmorales: should it really be only cassava_trait here? maybe make this a parameter to pass

nmenda: fixed 
d3720b64b8b19be433aab711f74edce33eea29e7

PR #1072, opened 2017-06-14
Topic/solgs/fast pca

isaak:
--initially created to implement the SNPRelate PCA for fast calc. But most of the code happened to be merged to another branch which is merged to master (forgot to document this during the merge).
--general code improvement
--tests pass
--turns off MAF filter
--uses Dataset genotype query for selection pops


lib/SGN/Model/solGS/solGS.pm
```diff
@@ -1171,14 +1155,26 @@ sub _create_genotype_row {
...
     }
 
     return $geno_values;
 
 }
 
+sub round_allele_dosage_values {
+    my ($self, $geno_values) = @_;
+
+    my $round = Math::Round::Var->new(0);
```
nickmorales: this might be slow to use new inside of the foreach my $marker (@markers) loop?

isaak: Indeed slow... on another on going branch I will look into doing the rounding in R (where there should be faster methods). 

In the interest of other applications/users that would use the genotype data, alternatively loading rounded dosage values to the database would be a solution.

PR #1166, opened 2017-07-18
Topic/dt feature search

nmenda:
rewrote feature search with dataTables
This replaces old feature search which does not work anymore, but eventually will keep only the gene/locus search tab and retire completely the feature table search


lib/SGN/Controller/Search/Feature.pm
```diff
@@ -37,11 +46,13 @@ sub oldsearch : Path('/feature/search') Args(0) {
 
 sub search :Path('/search/features') Args(0) {
```
nickmorales: There's two /search/features paths? (the others in SGN::Controller::Feature above). maybe one should be deprecated

mason/search/features.mas
```diff
@@ -1,33 +1,138 @@
 <%args>
-  $maximum_export_size
+ $organisms         => undef
+ $feature_types     => undef
 </%args>
 
 <& /util/import_extjs.mas &>
```
nickmorales: this mason (import_extjs.mas) is trying to pull in js and css that no longer exists. maybe remove this line and deprecate that mason file?

lib/SGN/Controller/Search/Feature.pm
nmenda: looks like the older one in SGN::Controller::Feature needs to be deprecated

mason/search/features.mas
nmenda: https://github.com/solgenomics/sgn/commit/e71aa45679806d6a7675f1c5985d9da175ef2af6

lib/SGN/Controller/Search/Feature.pm
nmenda: https://github.com/solgenomics/sgn/commit/3126858e6f2fc89a2371819bf7871e8d8a30f7a1

PR #1167, opened 2017-07-18
fixes whitespace mismatches in geno download and test

bellerbrock:


nickmorales:
wierd, this causes the test to fail for me, while the previous test worked.

PR #1194, opened 2017-08-01
Direct trial phenotyping.

aco46:


mason/site/header/head.mas
```diff
@@ -18,9 +23,13 @@
   <!-- <link rel="stylesheet" type="text/css" href="/css/dataTables.bootstrap.min.css"> -->
   <link rel="stylesheet" type="text/css" href="/css/buttons/buttons.bootstrap.min.css">
   <link rel="stylesheet" type="text/css" href="/css/bootstrap3-xl-class.css">
-  
+
+  <link rel="stylesheet" href="/css/bootstrap-responsive.min.css">
```
nickmorales: could you use the assets plugin to load these css files only on your page? that way there is no chance that these css classes interfere with other pages

static/css/bootstrap-responsive.css
```diff
@@ -0,0 +1,1109 @@
+/*!
```
nickmorales: we should already have bootstrap css so no need for this

static/css/bootstrap-responsive.min.css
```diff
@@ -0,0 +1,9 @@
+/*!
```
nickmorales: we already have boostrap css

static/css/bootstrap.css
```diff
@@ -0,0 +1,6158 @@
+/*!
```
nickmorales: again we already have bootstrap css

PR #1203, opened 2017-08-04
Topic/trial treatments

nickmorales:
Treatments can be applied to plots and plants. This application can happen during trial xls upload, during the trial creation, or from the trial detail page.

Phenotypes can be tied to treatments using xls spreadsheet or fieldbook


nickmorales:
will work on the split plot in this branch before merge

PR #1212, opened 2017-08-09
Topic/co 334

nmenda:
after merging need to update the local conf keys for trait_ontology_db_name and onto_root_namespaces


nickmorales:
one issue with this is if people are already collecting measurements using the CO terms, they will need to update their files to CO_334 before they can upload phenotypes

nmenda:
yes, they will . Maybe we can do a quick check on the cassava-users mailing list

PR #1232, opened 2017-08-15
Topic/interactive filtering

dauglyon:


lib/SGN/Controller/AJAX/BreedersToolbox/GraphicalFilteringAJAX.pm
```diff
@@ -0,0 +1,215 @@
...
+sub common_traits_by_trial_list_GET : Args(0) {
+   my $self = shift;
+   my $c = shift;
+
+   #get list ID from url param
+   my $trial_list_id = $c->request->param('trial_list_id');
+
+   #get userinfo from db
+   my $schema = $c->dbic_schema("Bio::Chado::Schema");
+   my $user = $c->user();
```
nickmorales: you should check that the user is logged in first on these ajax requests

dauglyon: - [x] [added check and alert if user is not logged in upon plot phenotypes AJAX call](https://github.com/solgenomics/sgn/pull/1232/commits/ce2ae21c95a88eab35f87ae8af51db33d21043cf)

PR #1239, opened 2017-08-23
Expands manage locations page to include map and edit/upload features

### LINKED ISSUE #1003: Add ability to edit a location after it's creation
For example, to allow the addition of long/lat and altitude info, or an edit to the name.

bellerbrock:
closes #1003


db/00083/UpdateGeolocationCvName.pm
```diff
@@ -0,0 +1,77 @@
...
+    print STDOUT "\nChecking if this db_patch was executed before or if previous db_patches have been executed.\n";
+
+    print STDOUT "\nExecuting the SQL commands.\n";
+
+    my $schema = Bio::Chado::Schema->connect( sub { $self->dbh->clone } );
+
+    my $row = $schema->resultset("Cv::Cv")->find( { name => 'geolocations_property' } );
+
+    if (defined $row) {
+        $row->name('geolocation_property');
```
nickmorales: some databases already have a cv called geolocation_property along with geolocations_property (e.g. cassavabase) so the patch will fail on them. in cassavabase there is one cvterm under the geolocation_property

system_cvterms.txt
```diff
@@ -1,207 +1,207 @@
 cv name	cvterm name	definition	date	person	dbpatch #	comment
-calendar	Fertilizer Event				00052
```
nickmorales: this file is a hard tab separated file. atom's whitespace add-on is clearing the end tabs. i would just disable the whitespace add-on so that it does not do these kinds of edits

mason/site/header/head.mas
```diff
@@ -18,6 +18,9 @@
   <!-- <link rel="stylesheet" type="text/css" href="/css/dataTables.bootstrap.min.css"> -->
   <link rel="stylesheet" type="text/css" href="/css/buttons/buttons.bootstrap.min.css">
   <link rel="stylesheet" type="text/css" href="/css/bootstrap3-xl-class.css">
+  <link rel="stylesheet" type="text/css" href="/css/leaflet.css">
```
nickmorales: we can load these css files specifically on the locations page like https://github.com/solgenomics/sgn/blob/master/lib/SGN/Controller/BreedersToolbox/TrialComparison.pm#L19

db/00083/UpdateGeolocationCvName.pm
bellerbrock: uhoh, I do see that now. Any thoughts on the best solution? I suppose I could modify the patch so that it checks for both possible cv names and merges them into a single 'geolocation_property' cv

system_cvterms.txt
bellerbrock: good catch! I had no idea that was enabled by default, just disabled it. I'll checkout the copy from master and redo the edits.

mason/site/header/head.mas
bellerbrock: Got it, that does make more sense.

db/00083/UpdateGeolocationCvName.pm
nickmorales: @bellerbrock if the patch could merge them into geolocation_property that would be great. as far as i know there is just that "geolocation address" cvterm that is a problem right now

bellerbrock: cool. commit 08d0d383859c712f27914fd01847435be6f95c23  updates the patch to handle a merge if necessary. anything else to address before merging?

PR #1254, opened 2017-08-29
Topic/solgs/update geno data filter

isaak:
--Improves correlation calc speed
--uses clone averages instead of BLUP(E)s for correlation calc
--faster  allele dosage rounding
--uses updated genoDataFilter R package

--requires upgrading genoDataFilter R package where ever solGS runs


nickmorales:
@isaak what would be the best way to upgrade to upgrade the genoDataFilter R package

PR #1262, opened 2017-08-31
adds list validation for a more informative error message during fieldbook trait file creation

### LINKED ISSUE #988: more informative error message during fieldbook trait file creation
if trait list has a term that does not exactly match TERM|CO:0000001, for example TERM| CO:00000001 then it does not give a good error message.
should use the trait list validation first

bellerbrock:
closes #988


lib/CXGN/List/Validate/Plugin/Traits.pm
```diff
@@ -21,26 +21,13 @@ sub validate {
...
-        if ($accession) {
-            $accession =~ s/\s+$//;
-            $accession =~ s/^\s+//;
-        }
-        if ($db_name) {
-            $db_name  =~ s/\s+$//;
-            $db_name  =~ s/^\s+//;
-        }
+
+        my ($trait_name, $full_accession) = split (/\|/ , $term);
```
nickmorales: This will not validate postcomposed terms properly (since they have | in them)

bellerbrock: ah, true, I forgot that composed terms now have additional | delimiters. Looks like the other parts of the Fieldbook object didn't account for that either. e2a9ec4a4e91eb4fbd610413a4a54eff44e81858 adds changes to handle the extra delimiters in composed traits. And removes some of the checks in the validation that are no longer needed, now that composed traits are also stored permanently with VARIABLE_OF relationships

PR #1282, opened 2017-09-13
Topic/accesions to seedlot tool

dauglyon:
This branch consists of two main additions:

### The Available Seedlots List Tool
This is the tool shown in the gif above. It allows one to see all seedlots related to a list of accessions and select which of those seedlots one would like to add to a new seedlot list. It implements this functionality using `mason/tools/available_seedlots.mas`.

### The Available Seedlots Mason File  
(`mason/tools/available_seedlots.mas`)  
This mason file, when included, adds a form element `#available-seedlots`. It also adds an object to the global JS namespace (`window.available_seedlots`). Passing a list of accession names (unique or synonyms) to `window.available_seedlots.build_table(names)` will populate the table inside the form with check boxes like seen in the video above. Calling the function `window.available_seedlots.get_selected()` will return a jQuery serialized form object for all the ticked boxes. The form table can be redrawn simply with another call to `window.available_seedlots.build_table(names)`.

### DIFF
--- a/js/CXGN/List.js
+++ b/js/CXGN/List.js
@@ -753,10 +753,10 @@ CXGN.List.prototype = {
         }
     },
 		seedlotSearch: function(list_id){
+			var self = this;
 			jQuery('#working_modal').modal('show');
 			var data = this.getListData(list_id);
 			var accessions = data.elements.map(function(d){return d[1];})
-			console.log(accessions);
 			if (window.available_seedlots){
 				window.available_seedlots.build_table(accessions);
 			} else {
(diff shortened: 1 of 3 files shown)


js/CXGN/List.js
```diff
@@ -753,9 +753,10 @@ CXGN.List.prototype = {
         }
     },
 		seedlotSearch: function(list_id){
+			var self = this;
 			jQuery('#working_modal').modal('show');
-			var accessions = this.getList(list_id);
-			console.log(accessions);
+			var data = this.getListData(list_id);
```
nickmorales: should use the getList function instead of getListData if you are just getting the contents

nickmorales:
Also, there is a bug that happens when i try to create a list from the selected seedlots: depending on how many times the "available seedlots" modal has been opened (e.g. i clicked "See available seedlots" button 3 times) that defines how many times it will try to create the new seed lot list (this results in errors because the same list name cannot be created 3 times). You may need to move your js out of available_seedlots.mas to fix this

PR #1292, opened 2017-09-18
Origin/topic/banana analysis pipeline

bauchetg:
R scripts to run the banana analysis pipeline and report creation by M.Karanja


bauchetg:
@nickmorales Sorry for confusion and thanks for pointing out this, I moved scripts one level up. It should be fine now.

PR #1316, opened 2017-09-29
More stringent checks of user defined names, and script to fix existing names with extra whitespace

### LINKED ISSUE #1080: Some trial names have whitespace at the end which causes issues with lists
For example, 13ayt25pdUB, 13grcUB, 13NEXTGenC0MK, 13pyt28highflowerIB, 13pytcrossingblkUB, 13uyt15yrt8AUB, and 13uyt25sgMK.

Adding them to a list in the search wizard works, but then the list does not validate because in the db they have whitespace at the end. A similiar problem still exists for some stocks.

bellerbrock:
Closes #1080


lib/CXGN/Trial/TrialCreate.pm
```diff
@@ -259,9 +266,9 @@ sub save_trial {
 		};
 	}
 	if ($error) {
-		return ( error => $error );
+		return { error => $error };
 	}
-	return ( trial_id => $project->project_id );
+	return { trial_id => $project->project_id };
```
nickmorales: returning a different data structure has effects on code that uses this, so the tests don't pass anymore

bellerbrock: true, thanks for checking. Latest commit should fix this.

PR #1339, opened 2017-10-09
Fixed long-url error issue with the availible_seedlots table for long lists

dauglyon:
Switched the availible_seedlots AJAX method from GET to POST and move the list of names from the URL to the POST body.


dauglyon:
@nickmorales Finally fixed this, it got lost in my to-do list, apologies.

PR #1496, opened 2017-12-04
Topic/upload plot gps

### LINKED ISSUE #1192: storing GPS coordinates at trial plots level
@bellerbrock some bullet points from discussion
-GPS coord could be added at a plot level (for ex for spatial analysis, by linking lat/long to row/cols data).
-Figure out if fieldbook can actually transfer such data.
-In this case infos could be linked to stockprops.

nickmorales:
Description 
Allows user to upload an xls file with WGS84 coordinates.
These coordinates are then saved in a GEOJSON string as a stockprop of the plot.

CXGN::Trial::TrialLayout then includes this in its cached json string, making it easy for someone to call a query for all trials in a location and then display the coordinates for those trial layouts on a map.

closes #1192

Checklist 
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [x] New feature
  - [x] User-Facing Change
    - [x] The user manual in `/docs` has been updated.


lib/SGN/Controller/AJAX/TrialMetadata.pm
```diff
@@ -639,6 +639,169 @@ sub trial_upload_plants : Chained('trial') PathPart('upload_plants') Args(0) {
...
+    my $upload_plot_gps_txn = sub {
+        my %plot_stock_ids_hash;
+        while (my ($key, $val) = each(%$parsed_data)){
+            $plot_stock_ids_hash{$val->{plot_stock_id}} = $val;
+        }
+        my @plot_stock_ids = keys %plot_stock_ids_hash;
+        my $plots_rs = $schema->resultset("Stock::Stock")->search({stock_id => {-in=>\@plot_stock_ids}});
+        while (my $plot=$plots_rs->next){
+            my $coords = $plot_stock_ids_hash{$plot->stock_id};
+            my $geo_json = {
```
dauglyon: Is the same information simply being stored in both coordinate systems here, or can the user set two different polygons in the different systems? The later seems dangerous. And if it's the former, wouldn't it be better to convert to a single internal representation and then convert out of that representation when needed? It would halve the data stored per plot.

nickmorales: right it is saving both kings of coordinate system in the geo json string. i didn't have time to check how to convert one to the other, but i know that the users like UTM31N because its based in meters, but i think we want WGS84 to really display on maps. i'm not sure so i just saved both.

dauglyon:
Maybe it would be better to just have one or the other stores then? Then, once conversion is worked out, all the non-gps entries can be converted. It doesn't seem good to have two representations of the same thing that can be set independently. I can take a look into conversion if you want

nickmorales:
it would probably be better to only store one format. if you can that would
be great

dauglyon:
Here is a module for UTM conversion. It looks like currently you have hardcoded zone 31N in for the UTM coordinates but this only account for plots in this region: 
screen shot 2017-12-04 at 11 45 05 am
So, there should be a field to indicate zone as well as UTM coords for that location entry method.

nickmorales:
@dauglyon maybe it is better only to allow upload of WGS84 and only store WGS84. What do you think?

dauglyon:
@nickmorales As long as it wouldn't create too much of a barrier to entry to the breeders uploading the data, that makes sense to me.

PR #1741, opened 2018-03-20
Topics/workflows

dauglyon:
Description 

Adds a mason component for creating workflows!

Also adds *.org/styleguide which is a place put put examples of mason component usage for commonly used components. 

Checklist 
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [x] New feature
  - [ ] User-Facing Change


dauglyon:
@bauchetg This may be useful for addressing the workflows you identified with those graphs you showed in lab meeting

bauchetg:
@dauglyon this looks great, indeed it will be very useful for these processes.
Is this something you see as part of the https://solgenomics.github.io/sgn/ ? 
Or integrated more directly to the database (like a workflow section?)
As well this type display seems a great fit for activity tracking:
 https://github.com/solgenomics/sgn/projects/13

dauglyon:
@bauchetg Definitely something to be integrated directly into the DB rather than in the documentation. The idea is to have predefined "workflows" where each step is one of the nodes in the graphs you showed. Then, alternately have the functionality at each of those steps also available independently.

for example,

we might have a workflow at `/workflows/upload_accessions` with steps to upload an accession, add its pedigree, etc. 

but we would also have individual pages for doing those same things. Ideally they would both be wrappers for the same mason file.

bellerbrock:
@dauglyon Wow, great work! This will be incredibly useful. 

And solid demo too, were you inspired by Xzibit? Now it just needs a "Yo dawg, I heard you like workflows . . ."

dauglyon:
26t7j8

PR #1774, opened 2018-03-30
Topic/solgs/combining trials fix

isaak:
-- fixes combining data from certain datasets

### DIFF
--- a/R/solGS/combine_populations.r
+++ b/R/solGS/combine_populations.r
@@ -4,13 +4,12 @@
 
 options(echo = FALSE)
 
-library(stats)
 library(stringr)
 library(randomForest)
-library(plyr)
 library(lme4)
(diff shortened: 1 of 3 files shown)


R/solGS/combine_populations.r
```diff
@@ -4,13 +4,13 @@
...
 
-library(stats)
 library(stringr)
 library(randomForest)
-library(plyr)
 library(lme4)
 library(data.table)
 library(phenoAnalysis)
+library(dplyr)
+library(tibble)
```
nickmorales: is tibble is the R deps?

isaak: Yes, it is. But it is already installed in the systems; it is a dependency of another package that is listed in the R files.

PR #1815, opened 2018-04-12
Topic/guided help

nickmorales:
Description 
Adds "genotyping trial upload" to homepage "guided helps".

Checklist 
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


bauchetg:
@nickmorales as we discussed
Step 4 (trial linkage) needs to be refined according to source info.
if tissue, plant or plot barcode info no need for step 4, if accession barcode info then it should be maintained.

nickmorales:
@bauchetg the linkage happens behind the scene now when a genotyping trial is added, so i cleaned up the add/upload trial and add genotyping trial interfaces

PR #1836, opened 2018-04-17
Topic/redesign stock detail page

nickmorales:
Description 
Redesigns stock detail page to look like the new trial detail page.
Has all the same sections as before.

There is a "phenotype section" that i could not account for at the bottom of the page. Is this the same as the "traits assayed section"?

Checklist 
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


bauchetg:
There is a "phenotype section" that i could not account for at the bottom of the page. Is this the same as the "traits assayed section"?
The trait section and phenotype section seem to be the same with slight difference in the output, like in https://cassavabase.org/stock/39671/view
I would remove the pheno section

PR #1940, opened 2018-06-11
update phenotype data structuring..

isaak:
update phenotype data structure; filter out newly added metadata columns


isaak:
This needs to be merged along with the new phenotype search update.

PR #1949, opened 2018-06-18
upload of coordinate custom genotyping trial template

### LINKED ISSUE #1865: Need additionnal custom template upload from Coordinates
Expected Behavior 
Coordinate PhenoApp offers preset template such as "DNA Plate" (already handled in cassavabase) with this format: date | plate_id | plate_name | sample_id | well_A01 | well_01A | tissue_id | dna_person | notes | tissue_type | extraction
It can be accessed using "I am uploading a Default Coordinate Android app plate design"

However the template  "DNA plate" sets randomly the control well which can be a problem for some genotyping vendor. This can be overcome by setting a custom template using "New template", where the default format is:
Value | Column | Row | Identification | Person | Date
Where fields are:
-Value: store the barcode content (here "plot name")
-Column: column number (integer)
-Row: row number (integer, not letter)
-Person: same as "dna_person"
-Date:  same as date in "DNA Plate" format (YYYY-MM-DD)

We would need to add this format under a second option like: "I am uploading a Custom Coordinate Android app plate design".
Additional options are offered (adding optional field, exclude field, or custom naming) but these would add more complexity.

Hello @trife is there any major changes planned in Coordinates output formats  we should keep in mind for further compatibility with cassavabase?
From https://github.com/PhenoApps/Coordinate it seems that some multiple plate could be exported in a single file, is that correct?

For Bugs:
---------

nickmorales:
Description 
Can upload a CSV file with the following format:
Value | Column | Row | Identification | Person | Date

This is the default "Custom" template from Coordinate app

closes #1865 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/js/CXGN/BreedersToolbox/GenotypingTrial.js
+++ b/js/CXGN/BreedersToolbox/GenotypingTrial.js
@@ -76,7 +76,12 @@ jQuery(document).ready(function ($) {
         if (uploadFileXLS === ''){
             var uploadFileCoordinate = jQuery("#genotyping_trial_layout_upload_coordinate").val();
             if (uploadFileCoordinate === ''){
-                submit_genotype_trial_create(plate_data);
+                var uploadFileCoordinateCustom = jQuery("#genotyping_trial_layout_upload_coordinate_template").val();
+                if (uploadFileCoordinateCustom === ''){
+                    submit_genotype_trial_create(plate_data);
+                } else {
+                    submit_genotype_trial_upload(plate_data);
+                }
             } else {
                 submit_genotype_trial_upload(plate_data);
             }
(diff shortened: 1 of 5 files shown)


bauchetg:
@nickmorales just tested the custom form, it loads well. However there may be a problem regarding the column-row info load and convertion.
 Currently custom form only takes integers for both col and row (not letter for row as in standard template). It should take the letter option but Trevor mentioned it currently not functional.
So couple of problems:
1-The test file load goes through but while on the genotyping page, the plate display doesn t indicate the collected well. 
2-The "intertek form download" is incorrect in the "well location" info field. 
3-Same for the "plate layout download" in the "plot number field".
We need to add a Column integer convertion to letter like:
{1=a,2=b,3=c,4=d,5=e,6=f,7=g,8=h}

nickmorales:
I can just restrict the Column to be a letter. that would solve all those
issues

bauchetg:
Right but unfortunately in the custom form Coordinates use only integers for both row and column (#1865) as mentionned. In its next release it will allow both int and letter for Rows.

PR #1997, opened 2018-06-29
Topic/materialized subquery fix

### LINKED ISSUE #1991: Materialized Views should use cv.name as well as cvterm.name to retrieve type_ids
Expected Behavior 
Encountered an error building materialized views in the new Dawson database - the subquery to return cvterm_id returned more than one row. Looks like it's caused by mutlitple cvterms of name 'plot', which in not an error in the database because they are used in different cvs. Meaning the subqueries should also specify the cv name that they search in. 

For Bugs:
---------

bellerbrock:
Description 
Closes #1991 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


db/00099/FixMatviewSubqueries.pm
```diff
@@ -0,0 +1,1405 @@
...
+
+
```
nickmorales: shouldn't the seedlots views be dropped before recreating as well?

bellerbrock: Yes, good point. It worked when I tested it, probably because the materialized_phenoview drop cascades to the seedlot views. But it's better to make the drops explicit. Should be fixed now.

nickmorales:
@bellerbrock is this still relevant?

bellerbrock:
Hi @nickmorales, I think so, though I'd forgotten about it. The addition of cv names was an important fix for some databases that had cvterms with identical names in different cvs. 

That said it may be better to abandon this branch and just include the cvname fix in a larger patch that drops the materialized_phenoview table and migrates the binary views to use the materialized_phenotype_jsonb_table. That's the most important improvement to include in any updates I think

bellerbrock:
Closing this for now

PR #2120, opened 2018-08-21
Set fieldbook barcode format as default on barcode generation page.

aco46:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/mason/barcode/index.mas
+++ b/mason/barcode/index.mas
@@ -159,7 +159,7 @@ my $date = join "/", ($dt->year, $dt->month, $dt->day);
               jQuery("#2D_fieldbook_option").show();
           }
           else if (barcode_type == "1D") {
-            jQuery("#2D_fieldbook_option").hide();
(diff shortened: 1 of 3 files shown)


bauchetg:
Thanks Alex, seems the "print duplicate labels per row option" is always checked, whatever the option selected, do we need it?

aco46:
Yes, it's always checked. It is different though for each printing format. Each printing format has a specified number of duplicate per row that goes along with the printing material.

bauchetg:
@aco46 sorry you are right I actually missed the "customed barcode" option (it is one level up). It is great to have it!
Having the custom option within each preset label types (plot/plant/etc...) would be helpful
A nice feature would be to have the same trial selected in the label as the trial page one you call it from. Maybe this is a little detail oriented :-) .

PR #2156, opened 2018-09-07
Topic/comments as notes

### LINKED ISSUE #745: Add a comment 'trait'
NCSU is interested in adding a comment 'trait' that is separate, but phenotyped using the fieldbook liek traits in the ontology. It would get stored as the plot description, and plot descriptions could all be pulled into the accession descriptions on the accession detail page.

bellerbrock:
Description 

Allows upload of notes/comments as notes stockprop. Closes #745

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lib/CXGN/Phenotypes/StorePhenotypes.pm
```diff
@@ -557,6 +566,21 @@ sub store {
...
+    my $operator = shift;
+    my $note = $note_array->[0];
+    my $timestamp = $note_array->[1];
+    $operator = $note_array->[2] ? $note_array->[2] : $operator;
+
+    print STDERR "Stock_id is $stock_id and note in sub is $note, timestamp is $timestamp, operator is $operator\n";
+
+    $note = $note ." (Operator: $operator, Time: $timestamp)";
+    my $stock = $self->bcs_schema()->resultset("Stock::Stock")->find( { stock_id => $stock_id } );
+    $stock->create_stockprops( { 'notes' => $note }, { autocreate => 1 } );
```
nickmorales: i would avoid using the autocreate here. you could add a db patch for 'notes' as 'stockprop'

nickmorales: right, i wouldn't advocate for the use of that there! i think we decided a long time ago that cvterms would only be created via db patches and not in the code

bellerbrock: Haha ok guess I wasn't thinking about that and took a page out of the wrong book. Or a line out of the wrong module.

The notes stockprop is added in db patch 88, so I just removed the unnecessary autocreate. We should probably should refactor how add stockprops is handled on stock detail pages but not trying to tackle that here.

nickmorales:
nice way to handle the non-ontology term "notes" as a stockprop, while enabling it to be used during phenotype upload and download

Issue #28, opened 2011-09-09
Finish organizing test suite files


leto:
Too messy right now!

PR #2217, opened 2018-10-12
Topic/solgs/clustering

isaak:
-- adds K-means based clustering feature as standalone and component of the trial detail page and solGS workflow
-- general code improvement


isaak:
this goes with  the Topic/solgs/clustering #28  cassava repo branch

PR #2266, opened 2018-11-11
Topic/solgs/anova update 031118

isaak:
-- adjusted means call can take now model output argument. 
-- This optional argument improves speed
-- general code improvement


isaak:
-- requires latest phenoAnalysis (0.1.6. 4) R package install

PR #2304, opened 2018-12-12
Topic/stock obsolete catalyst

nickmorales:
Description 
Making a stock obsolete from the stock detail page now uses a catalyst rest controller instead of the cgi-bin script.
Also, adding a population (on the manage accessions page) now works with a list of accessions that contains synonyms.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [x] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Stock.pm
+++ b/lib/CXGN/Stock.pm
@@ -282,9 +282,9 @@ sub store {
                 $self->_store_stockprop('organization', $self->organization_name());
             }
             if ($self->population_name){
-		print STDERR "**STOCK.PM This stock has population name " . $self->population_name . "\n\n";
-		#DO NOT INSERT POPULATION RELATIONSHIP FROM THE STOCK STORE FUNCTION
-		##$self->_store_population_relationship();
+                print STDERR "**STOCK.PM This stock has population name " . $self->population_name . "\n\n";
+                #DO NOT INSERT POPULATION RELATIONSHIP FROM THE STOCK STORE FUNCTION
+                $self->_store_population_relationship();
             }
 
         }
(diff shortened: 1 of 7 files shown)


lib/CXGN/Stock.pm
```diff
@@ -282,9 +282,9 @@ sub store {
                 $self->_store_stockprop('organization', $self->organization_name());
             }
             if ($self->population_name){
-		print STDERR "**STOCK.PM This stock has population name " . $self->population_name . "\n\n";
-		#DO NOT INSERT POPULATION RELATIONSHIP FROM THE STOCK STORE FUNCTION
-		##$self->_store_population_relationship();
+                print STDERR "**STOCK.PM This stock has population name " . $self->population_name . "\n\n";
+                #DO NOT INSERT POPULATION RELATIONSHIP FROM THE STOCK STORE FUNCTION
+                $self->_store_population_relationship();
```
nmenda: I think we need to remove this from the store function and explicitly call store_population_relationship from the AddPopulation object. This side effect can store unwanted stocks and stock_relationships unintentionally

lib/CXGN/Stock.pm
```diff
@@ -309,9 +309,9 @@ sub store {
             $self->_update_stockprop('organization', $self->organization_name());
         }
         if ($self->population_name){
-	    print STDERR "**STOCK.PM This stock has population name " . $self->population_name . "\n\n";
+            print STDERR "**STOCK.PM This stock has population name " . $self->population_name . "\n\n";
             #DO NOT INSERT POPULATION RELATIONSHIP FROM THE STOCK STORE FUNCTION
-            ##$self->_update_population_relationship();
+            $self->_update_population_relationship();
```
nmenda: move this out of the store function in Stock.pm 
Call the update from AddPopulation

nickmorales:
I agree we should move that into a population related object. If cxgn::stock is used properly it will not add those relationships but I agree it is a risk. This was needed for the tests to continue to all pass but we should think about refactoring

PR #2332, opened 2019-01-24
Topic/case insensitive accessions

nmenda:
Description 

when storing a new trial design the accession lookup should be case insensitive 
I did not modify the lookup for plot names, seedlots, etc. 
Could do too if we decide this is something we want to add.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/BreedersToolbox/StocksFuzzySearch.pm
+++ b/lib/CXGN/BreedersToolbox/StocksFuzzySearch.pm
@@ -49,7 +49,7 @@ sub get_matches {
     my %results;
     my $error = '';
     print STDERR "FuzzySearch 1".localtime()."\n";
-
+    
     my $synonym_type_id = SGN::Model::Cvterm->get_cvterm_row($schema, 'stock_synonym', 'stock_property')->cvterm_id();
     my $stock_type_id = SGN::Model::Cvterm->get_cvterm_row($schema, $stock_type, 'stock_type')->cvterm_id();
     my $q = "SELECT stock.uniquename, stockprop.value, stockprop.type_id FROM stock LEFT JOIN stockprop USING(stock_id) WHERE stock.type_id=$stock_type_id";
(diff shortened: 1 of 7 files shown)


nickmorales:
I'm not sure if we should encourage people to use the wrong names. For
example: 'accessionmUSA' and 'accessionMUSA' are clearly different in
meaning and could be confused in a case-insensitive match.
Also if a user has case-insensitive accession names in their own files, it
can mess up their own analysis if we are changing names behind the scenes.

nmenda:
@nickmorales the stock search is case insensitive (for searching and storing new accessions) , so the accession lookup when uploading a new trial layout should probably also be case insensitive.
If we keep it case sensitive then should modify the error message instead of 'accession not found'

bauchetg:
If I may chime in, both of you have a point. Having a case insensitive search is great for all legacy data/material with non structured named. The current case sensitive entry generates duplicates in some systems not in others, mainly where no naming conventions are adopted by the community. I would favor the case insensitive solution so it encourages such approach. Below a concrete case in Musa
screen shot 2019-01-25 at 12 18 09 pm
To solve it all, could case sensitivity be a local conf param?

titima15:
@nmenda @nickmorales During training, we always emphasize that stock names are case sensitive and it is easy to check if there is any doubt about uppercase/lowercase by using "Fuzzy Search". It will be very confusing if we allow case insensitive.

nmenda:
@bauchetg that screenshot is painful! 

@titima15 @bauchetg my point is that if someone tries to upload a file with a case insensitive accession name then it would be treated as an existing accession and would allow storing the layout.

If the accession in the database is named 'xyz'  and the layout file has the name 'XYZ' then the current error message just says 'accession name 'XYZ' not found.
The database field is varchar, which is case sensitive, but we do not allow storing case sensitive accessions. 
The accession search page is also case-insensitive

Alternatively, we could alter the error message and ask the user to modify the accession case to the one we use in the database.

I think either option is fine

nickmorales:
@bauchetg do you if those musa accessions were added through the web interface? when adding accessions using the web interface, if the accession is not an exact case sensitive match with a name already in the database, the fuzzy search does a case insensitive match across all names in the database and shows them to the user. the user can then add their name as a synonym to the existing accession, etc.

nmenda:
if we were to decide that all accession names in the database should be case insensitive, then the stock fuzzy search needs to be changed, move the case-insensitive lookup from 'fuzzy matches' to 'exact matches'  https://github.com/solgenomics/sgn/blob/master/lib/CXGN/BreedersToolbox/StocksFuzzySearch.pm#L83

PR #2336, opened 2019-01-30
fix list of trials arg bug..

isaak:


lukasmueller:
Can you provide more info on what this one if fixing?

PR #2342, opened 2019-02-08
Topic/solgs/cluster gebv

isaak:
-- clustering with phenotype, GEBVs and genotype
-- as standalone
-- integrated in the trial detail page, solGS workflow
-- uses Dataset, lists, trial, training datasets
-- general code improv't
-- bug fixes


isaak:
requires latest phenoAnalysis and genoDataFilter R packages

lib/SGN/Controller/solGS/pca.pm
```diff
@@ -160,64 +174,124 @@ sub pca_result :Path('/pca/result/') Args() {
...
+	    my $ret->{status} = undef;
+	    my $pca_scores    = $c->controller('solGS::solGS')->convert_to_arrayref_of_arrays($c, $pca_scores_file);
+	    my $pca_variances = $c->controller('solGS::solGS')->convert_to_arrayref_of_arrays($c, $pca_variance_file);
 
-}
+	    my $host = $c->req->base;
 
+	    if ( $host !~ /localhost/)
+	    {
+		$host =~ s/:\d+//;
```
lukasmueller: There should never be a check for localhost in the code, and I am not sure why this is needed in this case?

isaak: that is just for when working on my localhost.

lukasmueller:
What is different on your localhost that this is required? It should be the same, no?

lib/SGN/Controller/solGS/pca.pm
isaak: my localhost has the port number 3000 (http://localhost:3000).  On the test site and production site the 8080 port number is include when you do host and referer request, which creates links that break.

nickmorales:
I think the catalyst controller handles all these point of origin and payload steps. Maybe there is something special about this controller?

isaak:
Thanks, Nick. I removed the unnecessary base url coding. https://github.com/solgenomics/sgn/commit/e42c8684994eea6f1cefd111063eef584f9903ec

isaak:
Thanks, Nick. I removed the unnecessary base url coding. e42c868

PR #2412, opened 2019-04-09
Topic/genotypeprop nt order

nickmorales:
Description 
The NT key in genotypeprop is formatted in the order of Ref,Alt e.g. 'A,T' in the case where the separator is '/' in the VCF file and NT is unchanged when the '|' separator is used in the VCF file e.g. 'T|A'.
When uploading Intertek file formats the NT key in genotypeprop is ordered as Ref,Alt because these genotypes are always assumed unphased.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller:
Hi Nick,

I also have trouble following the code... Maybe we should have a quick meeting to discuss this?

PR #2501, opened 2019-06-06
Topic/phenotype deletion speed

### LINKED ISSUE #1414: trial phenotype deletion is very slow
Expected Behavior 
Even though deletion of all phenotypes in a trial is not common, it is very slow ~1 second/phenotype to delete on cassavabase. Would be more user friendly to speed this up.

For Bugs:
---------

### LINKED ISSUE #2494: GUI hangs while deleting phenotype sheet
Expected Behavior 

In Manage->Phenotyping.  I have a spreadsheet loaded for a trial.  The data is incorrect.  I would like to delete the spreadsheet (which I believe will delete the phenotypic data).  I expect to click "delete" next to the spreadsheet, and have some indication the delete is happening.  I expect this to happen in a timely fashion (not sure what "timely' should mean)

For Bugs:
---------

### Environment 

PanZeaBase at panzeabase.sgn.cornell.edu

#### Steps to Reproduce
At PanZeaBase:  go to Manage->Phenotyping.  Select the "delete" option for the uploaded file  2019-05-24_16:38:43_NAMjhBLUPS_phenotypeSheet.xls.  For over an hour, it shows "waiting for panzeabase.sgn.cornell.edu..."   I cannot tell if it is deleting data, or stuck trying to connect to the DB.  If I kill the window, the file still exists (has not been moved to obsolete list).  I refresh, try again - same behavior.

nickmorales:
Description 
Phenotype deletion now deletes nd_experiment entries in an asynchronous process, making the user experience fast.
Fast phenotype deletion using the following available methods:
1) Manage->Phenotypes: delete an uploaded file.
2) Trial detail page: delete trait data.
3) Phenotype Overwriting during Phenotype Upload
4) Using delete_trials.pl script

closes #2494 
closes #1414

Checklist 
- [x] Refactoring only
- [ ] Documentation only
- [x] Fixture update only
- [x] Bug fix
- [x] New feature
  - [ ] User-Facing Change


bin/load_phenotypes_spreadsheet_csv.pl
```diff
@@ -167,6 +168,11 @@ =head1 AUTHOR
 }
 
 my $store_phenotypes = CXGN::Phenotypes::StorePhenotypes->new(
+    basepath=>$opt_b,
+    dbhost=>$opt_H,
```
lukasmueller: This should not be part of the object, but done outside of the object and used to create a schema???

nickmorales: The reason I added those params is that now during phenotype deletion it runs:
my $async_delete = CXGN::Tools::Run->new();
$async_delete->run_async("perl $basepath/bin/delete_nd_experiment_entries.pl -H $dbhost -D $dbname -U $dbuser -P $dbpass -i $nd_experiment_ids");

The PhenotypeStore runs phenotype deletion in the case of overwriting previously loaded phenotype values, and so I needed those params in order to call the async_delete in the object.

Is there a better way?

lukasmueller: I guess the nd_experiment entries is the slow part to delete? That's why you factored it out to a script?

nickmorales: the main thing this PR does is factor out deletion of nd_experiment entries into an asynchronous process because it is extremely slow as the database become large

lukasmueller: Not sure if there is a better way...  let me sleep on it..

Another thing is that I completely re-factored CXGN::Trial in another branch, so we'll have to be careful when I merge that one...

Issue #2523, opened 2019-07-02
Set pedigree viewer to show only two generations by default


titima15:
Expected Behavior: Pedigree viewer should show only two generations and have an option to retrieve all pedigree info.
--------------------------------------------------------------------------

For Bugs:
---------

bauchetg:
@titima15 perhaps I missed something but why would you limit the pedigree depth to n = 2?

titima15:
@bauchetg (requested by Fola) Because when there are so many generations, it becomes very slow and large. Most of the time, breeders would like to see only 2 generations (or should we set for three generations). However, users have an option to expand the view.

aafolabi:
@bauchetg That is exactly what the problem is. Waiting time is about 1-2mins. Try and search "IBA070593". So if the default is set to 2 or maybe 3, the user can have an option to expand the view.
Thanks

bauchetg:
I see, this is a question for @dauglyon

titima15:
@bauchetg I assigned to Mirella because Lukas asked Mirella to handle d3 issue. It's a good idea that she consults David as well when needed.

PR #2563, opened 2019-07-31
tree draws just one parent instead the whole tree

MFlores2021:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ X] Bug fix
- [ ] New feature
  - [ ] User-Facing Change
#2523

### DIFF
--- a/js/source/legacy/brapi/PedigreeViewer.js
+++ b/js/source/legacy/brapi/PedigreeViewer.js
@@ -14,11 +14,13 @@ function PedigreeViewer(server,auth,urlFunc){
         var locationSelector = null;
         
         urlFunc = urlFunc!=undefined?urlFunc:function(){return null};
-                
-        pdgv.newTree = function(stock_id,callback){
+             
+        pdgv.newTree = function(stock_id,callback){   
             root = stock_id;
             loaded_nodes = {};
-            var all_nodes = [];
+            var all_nodes = []; 
+            var levels =0;
+            var number_ancestors =1;
             load_node_and_all_ancestors([stock_id]);
             function load_node_and_all_ancestors(ids){
                 load_nodes(ids,function(nodes){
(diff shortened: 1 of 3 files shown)


js/source/legacy/brapi/PedigreeViewer.js
```diff
@@ -14,11 +14,13 @@ function PedigreeViewer(server,auth,urlFunc){
...
+             
+        pdgv.newTree = function(stock_id,callback){   
             root = stock_id;
             loaded_nodes = {};
-            var all_nodes = [];
+            var all_nodes = []; 
+            var levels =0;
+            var number_ancestors =1;
             load_node_and_all_ancestors([stock_id]);
             function load_node_and_all_ancestors(ids){
```
dauglyon: `I would rename this function so it isnt misleading to load_node_and_ancestors`

js/source/legacy/brapi/PedigreeViewer.js
```diff
@@ -14,11 +14,13 @@ function PedigreeViewer(server,auth,urlFunc){
...
-                
-        pdgv.newTree = function(stock_id,callback){
+             
+        pdgv.newTree = function(stock_id,callback){   
             root = stock_id;
             loaded_nodes = {};
-            var all_nodes = [];
+            var all_nodes = []; 
+            var levels =0;
+            var number_ancestors =1;
```
dauglyon: `ancestor_depth`?

js/source/legacy/brapi/PedigreeViewer.js
```diff
@@ -14,11 +14,13 @@ function PedigreeViewer(server,auth,urlFunc){
         var locationSelector = null;
         
         urlFunc = urlFunc!=undefined?urlFunc:function(){return null};
-                
-        pdgv.newTree = function(stock_id,callback){
+             
+        pdgv.newTree = function(stock_id,callback){
```
dauglyon: If `number_ancestors`/`ancestor_depth` is added as an argument here, it can be controlled without modifying the source, in case someone wants deeper trees.

dauglyon:
@lukasmueller @MFlores2021 now that sgn can NPM, it might be better to make this change in @solgenomics/BrAPI-Pedigree-Viewer and bump the version, then include it as a dependency.

PR #2650, opened 2019-10-12
Update blast to new blast (>2.6)

lukasmueller:
Description 

Note that this branch requires the Debian blast package ncbi-blast+ to be installed. The legacy interface is not used anymore at all, so both the blast2 and ncbi-blast+-legacy package should not be required.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lib/CXGN/Blast/Parse/Plugin/#Basic.pm#
```diff
@@ -0,0 +1,25 @@
+
```
nickmorales: should this buffer (#Basic.pm#) be added?

lukasmueller: no... that should be removed. Good catch. Will do.

lukasmueller: Done!

lukasmueller:
I think you may have to re-review....

lukasmueller:
or not

PR #2671, opened 2019-10-24
Bbase-8 Convert Breedbase data types to BrAPI data types in variable pulls

ctucker3:
Description 
Added a function to convert breedbase datatypes to BrAPI data types. So far, converting numeric to Numerical, qualitative to Text, and if the trait has categories convert to Nominal. The trait is converted to whatever condition it passes first, the order is like so:

- Nominal check
- Text check
- Numerical Check

Breedbase can have any number of data types, but numeric and qualitative are the ones Nick Morales mentioned as being the two he has seen in the past. 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ x ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/BrAPI/v1/ObservationVariables.pm
+++ b/lib/CXGN/BrAPI/v1/ObservationVariables.pm
@@ -224,23 +224,8 @@ sub observation_variable_search {
 			}
 		);
 
-		# Convert our breedbase data types to BrAPI data types. If we find a type we want
-		# to convert, convert it. If there is a type, but we have no conversion for it,
-		# let it pass.
-		my $trait_format = $trait->format;
-
-		if (scalar @brapi_categories > 0) {
-			# If the trait has categories, convert to Nominal
-			$trait_format = "Nominal";
-		}
-		elsif ($trait_format eq "qualitative") {
-			# If the trait is qualitative convert to Text
-			$trait_format = "Text";
-		}
-		elsif ($trait_format eq "" || $trait_format eq "numeric" || ! defined $trait_format){
-			# If the trait is numeric or the data type is unspecified, convert to Numerical
-			$trait_format = "Numerical";
-		}
+		# Convert our breedbase data types to BrAPI data types.
+		my $trait_format = $self->convert_datatype_to_brapi($trait->format, scalar(@brapi_categories));
 
 		# Note: Breedbase does not have a concept of 'methods'.
 		# Note: Breedbase does not have a concept of 'scale'. The values populated in scale are values from cvprop.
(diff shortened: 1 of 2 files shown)


ctucker3:
[BBase-8 - GET variables - map breedbase datatypes to BrAPI datatypes](https://trello.com/c/8XGWBvbk/61-bbase-8-get-variables-map-breedbase-datatypes-to-brapi-datatypes)

lib/CXGN/BrAPI/v1/ObservationVariables.pm
```diff
@@ -345,5 +348,28 @@ sub observation_variable_detail {
...
+sub convert_datatype_to_brapi {
+	#If we find a type we want to convert, convert it.
+	# If there is a type, but we have no conversion for it, let it pass.
+	my $self = shift;
+	my $trait_format = shift;
+	my $num_brapi_categories = shift;
+
+	if ($num_brapi_categories > 0) {
+		# If the trait has categories, convert to Nominal
+		$trait_format = "Nominal";
```
nickpalladino: BrAPI datatypes:
- "Code" - This scale class is exceptionally used to express complex traits. Code is a nominal scale that combines the expressions of the different traits composing the complex trait. For exemple a severity trait might be expressed by a 2 digit and 2 character code. The first 2 digits are the percentage of the plant covered by a fungus and the 2 characters refer to the delay in development, e.g. "75VD" means "75%" of the plant is Crop Ontology & Integrated Breeding Platform Curation Guidelines 5/6/2016 9 infected and the plant is very delayed. 
- "Date" - The date class is for events expressed in a time format, e.g. yyyymmddThh:mm:ssZ or dd/mm/yy 
"Duration" - The Duration class is for time elapsed between two events expressed in a time format, e.g. days, hours, months 
- "Nominal" - Categorical scale that can take one of a limited and fixed number of categories. There is no intrinsic ordering to the categories
 - "Numerical" - Numerical scales express the trait with real numbers. The numerical scale defines the unit e.g. centimeter, ton per hectar, branches 
- "Ordinal" - Ordinal scales are scales composed of ordered categories 
- "Text" - A free text is used to express the trait.

Why nominal vs ordinal? Are the other brapi types handled correctly?

timparsons: My 2 cents...given that the values for ordinal/nominal are only a string value, it makes sense to assume ordinal, but if the list of categories gets changed to be a list of category objects (https://github.com/plantbreeding/API/issues/397), then there may be a reason to have them be separate

ctucker3: On Nick's point of all BrAPI types being handled, I'll address each one here for future reference. Code, Date and Duration, aren't handled in the conversion because the only data type formats that are used in Breedbase commonly (that we know of) are numeric and qualitative. 

Nominal and Ordinal has been discussed. 

Text is handled for the qualitative trait. But, if a trait has categories assigned, and has a data type of qualitative or numeric, it will be converted to Ordinal first. This takes into account the case when the trait data type is indicating whether the categories are numerical or qualitative. 

That being said, if a Breedbase data type is not explicitly converted in the conversion code, the data type that is listed in the database is passed in the response (if it is not blank, in which case it is converted to numeric). Since Breedbase allows for free text data types, this could allow Breedbase to store BrAPI data types in the future.

ctucker3: > My 2 cents...given that the values for ordinal/nominal are only a string value, it makes sense to assume ordinal, but if the list of categories gets changed to be a list of category objects ([plantbreeding/API#397](https://github.com/plantbreeding/API/issues/397)), then there may be a reason to have them be separate

I could see there being a reason that comes up to add the nominal data type conversion in the future and is a good thing to keep in mind for our v2.0 endpoints. We should be able to leave default conversion for categorical data to ordinal for now, since the v1.3 endpoint should always return category strings in the response.

PR #2681, opened 2019-11-05
Topic/solgs/pca test

isaak:
-- adds pca tests
-- sets timer for analysis status check duration.
-- fixes histogram data display
-- general code improv't


lukasmueller:
There are files of the form t/data/solgs/cache/03/92/9e5caf4641039352fb3821fb79a2ae68dd34 still in this commit... should they be there?

PR #2700, opened 2019-11-21
fixed derived trait to allow ontology database names with more than 2 caracters

laguem:
Description 
Derived trait module can now handle ontology database names with more than 2 characters. 
Also fixed accession validation in lists to be case insensitive and fixed a typo in a variable in StocksFuzzySearch.pm.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/List/Validate/Plugin/Accessions.pm
+++ b/lib/CXGN/List/Validate/Plugin/Accessions.pm
@@ -22,7 +22,7 @@ sub validate {
     my $h = $schema->storage->dbh()->prepare($q);
     $h->execute();
     while (my ($uniquename, $synonym, $type_id) = $h->fetchrow_array()) {
-        $all_names{$uniquename}++;
+        $all_names{lc($uniquename)}++;
         if ($type_id) {
             if ($type_id == $synonym_type_id) {
                 $all_names{$synonym}++;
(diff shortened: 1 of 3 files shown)


lib/CXGN/List/Validate/Plugin/Accessions.pm
```diff
@@ -22,7 +22,7 @@ sub validate {
     my $h = $schema->storage->dbh()->prepare($q);
     $h->execute();
     while (my ($uniquename, $synonym, $type_id) = $h->fetchrow_array()) {
-        $all_names{$uniquename}++;
+        $all_names{lc($uniquename)}++;
```
nickmorales: i'm not sure it is a good idea to change the user's input to lowercase here because accession names are implemented as case sensitive by the code, e.g. myaccession is different from MyAccession; however, you can have myaccession as a synonym to MyAccession and then you would not need to make this edit. What do you think?

laguem: I'll see what we could do with our accessions in our database, but having myaccession as synonym of Myaccession for all accessions in a database would slow down searches and increase the database size.  I understand some have accessions with similar names but different cases as different accessions.  Maybe we could have test with ucfirst (uppercase first) so myaccession = Myaccession but myaccession <> MyAccession ?
Like this:
    my @missing;
    foreach my $item (@$list) {
-        if (!exists($all_names{$item})) {
+       if (!exists($all_names{$item}) && !exists($all_names{ucfirst($item)})) {
            push @missing, $item;
        }
    }

Let me know what you think.

PR #2739, opened 2019-12-17
refactor smtp conf variables to include layer option and to match package syntax

### LINKED ISSUE #2738: Allow user to set smtp auth method
Expected Behavior 

At the moment [it is harcoded to ssl](https://github.com/solgenomics/sgn/blob/master/lib/CXGN/Contact.pm#L93), but some users may need to use tls.

For Bugs:
---------

bellerbrock:
Description 
closes #2738

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Contact.pm
+++ b/lib/CXGN/Contact.pm
@@ -85,6 +85,8 @@ sub send_email {
             my $smtp_port = $vhost_conf->get_conf('smtp_port');
             my $smtp_login = $vhost_conf->get_conf('smtp_login');
             my $smtp_pass = $vhost_conf->get_conf('smtp_pass');
+            my $smtp_auth = $vhost_conf->get_conf('smtp_auth');
+            my $smtp_from = $vhost_conf->get_conf('smtp_from') || $mailfrom;
 
             # If SMTP config values are found use external SMTP server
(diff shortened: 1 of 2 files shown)


bellerbrock:
> Can you please give default values for this to send as it used to?

@nickmorales if you mean Email::Send::SMTP::Gmail defaults, the docs say it uses the following if smtp, layer, or port aren't supplied.
```
smtp => smtp.gmail.com,
layer => tls,
port => 25
```
If login or pass aren't supplied then Contact.pm defaults to the old way of sending email via www_user@<hostname>

If you mean default values for the conf variables in local.conf then I suppose that depends on what the bulk of the databases will use. Any thoughts @lukasmueller? Can breedbase.org be configured as a send only smtp-server with a noreply@breedbase.org login? That could be logical default for all the breeding databases.

bellerbrock:
> With these changes, emails work with our SMTP server.
Awesome, thanks @dwaring87. @laguem also reports it works for him. Last thing is to fill in appropriate defaults. @nickmorales do you have any thoughts?

PR #2780, opened 2020-02-03
Topic/view controllers by name

dwaring87:
Description 

- Add a bug fix for the Locus view by name controller:
  - The `$c->detach()` function was throwing an error (`Couldn't detach to command "/locus/view/": Invalid action or component`)
  - Replaced with `$c->res->redirect(url, 301)`

- Add a function to view a Stock by name
  - Path: `/stock/view`
  - Query Params:
      - name = stock name
      - organism = (optional) organism abbreviation, genus, species or common name
  - If there are no matches, display an error with a link to the Stock search page
  - If there is 1 match, redirect to the stock detail page
  - If there are multiple matches, display a list of the matches with links to each stock detail page

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


dwaring87:
Hi @nmenda, I've added a function to the Stock controller to view a stock by name and organism.  Let me know what you think or if you have a better way to accomplish this.  Thanks!

lib/SGN/Controller/Locus.pm
```diff
@@ -93,22 +93,22 @@ sub view_by_name : Path('/locus/view/') Args(0) {
...
-    if ($locusname) { 
-	$locus = $c->stash->{locus} = CXGN::Phenome::Locus->new_with_locusname($c->dbc->dbh, $locusname);
+    if ($locusname) {
+        $locus = $c->stash->{locus} = CXGN::Phenome::Locus->new_with_locusname($c->dbc->dbh, $locusname);
     }
     
-    if (defined($locus->get_locus_id())) { 
-	my $locus_id = $locus->get_locus_id();
-	$c->stash->{locus} = $locus;
-	$c->detach("/locus/view/", [ $locus_id] );    #("/locus/$locus_id/view");
```
nmenda: where was this not working?

dwaring87: On our wheat instance: 
https://wheat.triticeaetoolbox.org/locus/view?symbol=1RS1AL&species=Wheat 

we get the error:
`[error] Couldn't detach to command "/locus/view/": Invalid action or component.`

nmenda: ok, I see. I don't think we use this piece of code. We use the locus name as the identifier in the URL 
https://solgenomics.net/locus/Solyc01g005000/view

https://github.com/solgenomics/sgn/blob/master/lib/SGN/Controller/Locus.pm#L261

dwaring87:
I don't know if we have a preference one way or the other, as long as an external site could get to the detail page of an item with just the name.  Is that what the detach is supposed to do, create a stable URL with the name?

nmenda:
I don't think we use the Locus view_by_name function

We use this for building a url with the locus genome identifier https://github.com/solgenomics/sgn/blob/master/lib/SGN/Controller/Locus.pm#L261

You could build a url with the stock name and organism
maybe something like stock/$uniquename/$organism
or stock/view_by_name/$uniquename/$organism

but I don't know if it's better than a redirect. I guess it depends on how you are planning on using it.

dwaring87:
I guess I found some old code first.  I'll take a look at that and see if I can set up something similar for stocks.  Having a stable URL might be better than a redirect if a link gets crawled by a search engine.

dwaring87:
Hi @nmenda - I just updated the Stock controller with two endpoints that use path parameters to view a stock by name:
  - **/stock/view_by_name/`$organism`/`$uniquename`** where `$organism` can be the organism abbreviation, genus, species or common name and `$uniquename` is the case sensitive stock unique name
  - **/stock/view_by_name/`$uniquename`** can be used when the organism doesn't need to be specified

If there is just 1 match, the stock detail page is displayed (not using a redirect).  If there are no matches, an error message with a link to the stock search page is displayed.  If there are multiple matches, an error message with links to stock detail pages of each of the matches is displayed.

lib/SGN/Controller/Stock.pm
```diff
@@ -286,6 +286,44 @@ sub view_stock : Chained('get_stock') PathPart('view') Args(0) {
...
+Path Params:
+    name = stock unique name
+
+Search for stock(s) matching the stock unique name.
+If 1 match is found, display the stock detail page.  Display an error for 
+0 matches and a list of matches when multiple stocks are found.
+
+=cut
+
+sub view_by_name : Path('/stock/view_by_name') Args(1) {
```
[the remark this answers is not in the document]
dwaring87: I think we would benefit from the lookups being case-insensitive, so I can change that.  I can also see using the same URL for both types of lookups being confusing.  I can change the one using the organism name to `view_by_organism`.  Thanks!

dwaring87:
@nmenda The last commit changes the stock and organism searches to be case-insensitive.  It also changes the URL for an organism and stock search to: **/stock/view_by_organism/`$organism`/`$stock`**

PR #2820, opened 2020-03-03
Topic/faster grm from wizard

nickmorales:
Description 
GRM download from the wizard was very slow when using yapri, but now is fast.
Uses rrblup to calculate GRM

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


nickmorales:
> The GRM is now super fast. The GRM matrix delivers stock id rather than stock names. Is it by design? Would be better to have stock names and/or tissue id

I think stock id is safer than using names because names can change, different synonym names, etc

PR #3111, opened 2020-10-21
Topic/genotype upload check protocol

### LINKED ISSUE #3107: Does breedbase validate vcf uploads against protocol specification?
So, when I upload a vcf with a new protocol (marker set) I am asked by the walk through to define a protocol. Then for subsequent uploads of vcfs genotyped with the same marker set I select the defined protocol. Currently, I validate the vcfs (marker order, ref/alt designation, marker number, info fields etc.) before I upload. I was wondering if breedbase performs any of these checks against the defined protocol? In other words, does breedbase validate vcf uploads against the protocol specification?

nickmorales:
Description 
Now when genotyping data is uploaded under the same genotyping protocol, the database will verify that the marker information is consistent (e.g. the alternate, reference, position, info, format, etc) for the genotyping protocol.

closes #3107 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Genotype/Search.pm
+++ b/lib/CXGN/Genotype/Search.pm
@@ -1195,17 +1195,12 @@ sub get_cached_file_search_json {
 
         open my $out_copy, '<', $tempfile or die "Can't open output file: $!";
 
-        if (!$self->forbid_cache()) {
-            $self->cache()->set($key, '');
-            $file_handle = $self->cache()->handle($key);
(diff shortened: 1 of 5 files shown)


lukasmueller (approved):
:-)

PR #3169, opened 2020-11-16
Topic/solgs/store gebvs

isaak:
-- organizes saved measurements analyses tools into one sub section
-- fixes kinship output file lookup (occurs when submitting the job)
-- improves x-axis labels visibility


lukasmueller (approved):
:-)

PR #3173, opened 2020-11-17
Topic/chromosome separated genotype storage

### LINKED ISSUE #3033: Split genotyping information into genotype props for different chromosomes
Expected Behavior 

Currently, the genotype prop represents an entire genome, which leads to issues when genome are large because large genomes may exceed the storage capacity of a jsonb field in postgres. The solution is to store each chromosome separately in a genotypeprop, which should be all associated with a single genotype entry.

For Bugs:
---------

nickmorales:
Description 
The genotype upload can now load many millions of markers without hitting the 1GB postgreSQL data cell limit. The genotypeprop and protocolprop entries are separated by chromosome.

This fixes tests in unit_fixture, unit, and unit_mech, most notably the t/unit_mech/brapi_v2.t test now runs against localhost and downstream tests were modified to account for the inserts and updates happening in the brapi tests.

closes #3033 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
:-)

PR #3177, opened 2020-11-19
Topic/dart gbs upload

isaak:
-- make stock name hash lookup case-insensitive
-- use sample type arg when loading into an existing protocol
-- scripts for pre-upload cleaning and structuring vcf files

-- requires Hash::Case::Preserve


lukasmueller (approved):
:-)

PR #3195, opened 2020-11-25
Changing from accession stock type to cross stock type

titima15:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/bin/change_accessions_to_crosses.pl
+++ b/bin/change_accessions_to_crosses.pl
@@ -2,7 +2,7 @@
 
 =head1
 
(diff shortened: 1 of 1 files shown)


titima15:
The script changes stock with type accession to type cross, then links each cross to parents, nd_experiment and a project

lukasmueller (approved):
Hi Titima, I would maybe add to the script documentation the exact use case when this should be run, because it is quite unusual?

:-)

PR #3200, opened 2020-11-30
add missing genotype protocol id arg

isaak:
-- fixes bug checking training pop size  with non-default genotyping protocols


lukasmueller (approved):
:-)

PR #3238, opened 2020-12-17
Topic/stock missing genotype

isaak:
-- hide 'Download all genotypes' on stock detail page
-- stop /stock/*/genotypes calls. (redirects to generic message page)


lukasmueller (approved):
:-)

PR #3258, opened 2021-01-04
Topic/dart gbs cleanup

isaak:
-- make hash table lookups case-insensitive
--  replace hard-coded stock type with db lookup


lukasmueller (approved):
:-)

Issue #3302, opened 2021-01-26
Enable test sites to send emails


lukasmueller:
Expected Behavior 
--------------------------------------------------------------------------

Emails sending is deactivated on test machines. For many purposes, such as demos, it would be useful for test sites to be able to send emails, so remove that limitation (email sending checks the production_server flag, so this check needs to be removed). 

For Bugs:
---------

PR #3312, opened 2021-01-30
Allow test sites to send emails (contact and registration verification emails)

lukasmueller:
Description 
This allows the test sites to also send emails, by removing the constraint that suppressed emails when production_server = 0.

Note that you still may have to 
* configure the correct github_access_token in sgn_local.conf (it can be copied from a production server), and that 
* the main_production_site_url, also in sgn_local.conf, needs to be set to the correct domain such that the confirmation link in the verification email goes to the correct site.
* set disable_emails = 0 in sgn_local.conf (this should be the default)
* postfix needs to be installed on the vm/server
* there might be additional firewall issues in sending mail

Closes issue #3302.

Checklist 
- [x] minor change
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Contact.pm
+++ b/lib/CXGN/Contact.pm
@@ -75,7 +75,7 @@ sub send_email {
 
     $body .= $request_info;
     print STDERR "$subject\n\n$body";
-    if ( $vhost_conf->get_conf('production_server') ) {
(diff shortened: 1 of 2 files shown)


titima15 (approved):
👍

PR #3387, opened 2021-03-11
add access token check and fix station search link

### LINKED ISSUE #3372: Location NOAA station edit reports invalid id when noaa_ncdc_access_token config variable is not set
Expected Behavior 
Tried saving the Clinton, NC station id `GHCND:USC00311881` in sweetpotatobase, but got the error message `The NOAA NCDC authorization token may not be valid! We could not verify your NOAA StationID. Please contact us!`

The id itself is from https://www.ncdc.noaa.gov/ and should be valid. [In the code](https://github.com/solgenomics/sgn/blob/7b5bcd94199385eea1503ee680958e31f9f53e42/mason/breeders_toolbox/location_map.mas#L423) it looks like the problem is that the `noaa_ncdc_access_token` is not set. It'd be better to check if the access token is defined before using it, and fail if it's not. It'd also be helpful to describe (maybe in the docs?) how to get/create a noaa access token.

For Bugs:
---------

bellerbrock:
Description 
Closes #3372 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


nickmorales (approved):
👍

PR #3416, opened 2021-03-22
Topic/seedlot upload error messages

dwaring87:
Description 

This improves the error messages returned when uploading seedlots:
- include the seedlot name with the error message
- check for undefined accession/cross id before creating a new seedlot
- check for errors returned from the seedlot store function

It also changes the add/upload buttons on the seedlot page to only be displayed when the user is logged in and displays a message to login if they are not.

There is still an issue that a seedlot file can pass validation but fail on storage if the accession name differs in case from the database entry (the validation is case-insensitive but the method to get the stock id from an accession name is case sensitive).

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
:-)

PR #3428, opened 2021-03-25
Topic/alter analysis comp trait handling

bellerbrock:
Description 

Adds missing owner_id to analysis trial create
Fixes analysis save errors by simplifying composed trait handling. 

@nickmorales the composed trait code was not working for me when trying to save mixed model analyses, but I'm not sure what other contexts it's used in and if these changes will break them. Could you help review?
In particular I'm not understanding the code written to handle composition of time terms - won't analyses only need to combine existing traits with a statistical ontology term?

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Analysis.pm
+++ b/lib/CXGN/Analysis.pm
@@ -17,9 +17,9 @@ The data structure is built using type ids that are different from a regular fie
 
 =item nd_experiment.type_id
 
-The nd_experiment.type_id links to 'analysis_experiment' (nd_experiment_property) (equivalent to 'field_experiment' in a trial), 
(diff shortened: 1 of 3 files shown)


nickmorales (approved):
@bellerbrock the idea was to allow someone to pass in an object for the traits, such as:
{
'Harvest index|CO_323:0000101' => ['day 10|TIME:0000102', 'additional info|ETC:0000202'],
}
and the system would post-compose the trait with the statistical ontology term and the additional terms.
This is useful for time-series terms that need to be post-composed with their time term.

PR #3433, opened 2021-03-28
Topic/fix download utf8

lukasmueller:
Description 
Fixes an issue with the download of pedigrees on the download page that contain characters in utf-8.
In essence, the read_file function of File::Slurp does not handle UTF-8 correctly, and should not be used in any code... (that was new to me too).

Thanks to Martin Lague for pointing out the issue.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


nickmorales (approved):
👍

PR #3468, opened 2021-04-09
Duplicate Genotypes Download

### LINKED ISSUE #3453: Dealing with duplicate genotypes
I was curious how BreedBase handles genotype data that is for the same accession on the same genotyping protocol. Every year we genotype all yield plots from the first rep of all trials and we genotype with the same SNP set. This generates lots of duplicate data for checks and accessions that were genotyped the previous year or the same year across trials.  Technically the lines are changing slightly each year with selfing and also we want to be able to track seedlot contamination/planting errors, so the duplicate data are useful but It doesn't seem like there is a way, on the user facing side at least, to manage the duplicate genotypes. Breedbase allows duplicates to be uploaded, but I'm not sure which version is being returned on download. It would be useful if there was a concept of a preferred duplicate that could be returned by default, but all duplicates were retained because we need a record for identifying mixups/contamination. Having the option to merge duplicates via consensus would also be useful.

nickmorales:
Option on wizard download to include duplicate genotypes in returned VCF or DosageMatrix. If the option is selected, the downloaded file will have the unique database genotype_id appended in the header columns.

Fixes bug in upload of duplicate genotypes in the same protocol. If duplicate genotypes were uploaded for the same stock in the same protocol, only the most recent genotypes were stored. Now the upload will actually store all the duplicate genotypes.

Description 

closes #3453 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/js/source/modules/wizard-downloads.js
+++ b/js/source/modules/wizard-downloads.js
@@ -6,14 +6,14 @@ import "../legacy/CXGN/Dataset.js";
  *
  * @class
  * @classdesc links to a wizard and manages showing relavant related data
- * @param  {type} main_id div to draw within 
- * @param  {type} wizard wizard to link to 
+ * @param  {type} main_id div to draw within
+ * @param  {type} wizard wizard to link to
  * @returns {Object}
- */ 
+ */
 export function WizardDownloads(main_id,wizard){
   var main = d3.select(main_id);
   var datasets = new CXGN.Dataset();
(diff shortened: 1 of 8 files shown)


lukasmueller (approved):
:-)

PR #3475, opened 2021-04-14
Footer: add git status information

dwaring87:
Description 

This adds git status information about the sgn repo to the footer, such as the latest commit hash, the latest tag, and the timestamp of the last commit.

This is added to the base mason component in /mason/site/footer/body.mas but will probably need to be incorporated into the site-specific mason file.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/mason/site/footer/body.mas
+++ b/mason/site/footer/body.mas
@@ -1,3 +1,8 @@
+<%init>
+  my $tag = `git describe --abbrev=0 --tags`;
+  my $commit = `git rev-parse --short HEAD`;
+  my $updated = `git log -1 --format=%cd`;
+</%init>
 
 <!-- BEGIN SITE SPECIFIC FOOTER. -->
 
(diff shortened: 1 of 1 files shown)


mason/site/footer/body.mas
```diff
@@ -1,20 +1,32 @@
+<%init>
+  my $tag = `git describe --abbrev=0 --tags`;
```
nickmorales, line 2: this is a great idea!

in the site specific masons it is possible to do something like (this would also work in this repo):
my $dir = $c->config->{basepath};
my $tag = `cd $dir; git describe --abbrev=0 --tags`;
my $commit = `cd $dir; git rev-parse --short HEAD`;
my $updated = `cd $dir; git log -1 --format=%cd`;

lukasmueller (approved):
:-)

PR #3521, opened 2021-05-12
Topic/refresh matviews

nmenda:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/bin/refresh_matviews.pl
+++ b/bin/refresh_matviews.pl
@@ -68,26 +68,28 @@ try {
     if ($mode eq 'fullview') {
 	@mv_names = keys ( %matviews );
 
-	print STDERR "*Setting currently_refreshing = TRUE\n";
-	my $cur_refreshing_h = $dbh->prepare($cur_refreshing_q);
-	$cur_refreshing_h->execute($state);
+
     } 
     if ($mode eq 'stockprop'){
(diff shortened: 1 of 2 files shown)


bin/refresh_matviews.pl
```diff
@@ -13,89 +13,115 @@ =head1 DESCRIPTION
...
+    }
+    
+    my $cur_refreshing_q =  "UPDATE public.matviews SET currently_refreshing=?";
+    my $state = 'TRUE';
+    
+    my @mv_names = ();
+    
+    if ($mode eq 'fullview') {
+	@mv_names = keys ( %matviews );
+	push (@mv_names, "materialized_stockprop", "materialized_phenotype_jsonb_table");
```
bellerbrock, line 70: Hey @nmenda , with the db patch update to the matviews table this line should no longer be needed, right?

Also I'm thinking we should add 'phenotype' and 'genotype' modes options here. But I see what you mean about needing many to many relationships to do it, maybe via a new matview_types table

nmenda: I removed this line and added a 'refresh_start' column to matviews. 
Will add more modes as we decide how to group the views in the db

lukasmueller (approved):
:-)

PR #3593, opened 2021-06-25
fix brapi data type check for search calls

ctucker3:
Fixes a bug where the brapi controller expects a json array body for search calls because its a POST call.


lukasmueller (commented):
:-)

lukasmueller (approved):

PR #3595, opened 2021-06-28
Topic/seedlots create from trial

### LINKED ISSUE #3587: Feature Request: Create Seedlots from Phenotyping Trial
Expected Behavior 

As requested by a breeder:

> I am trying to use the seed lot management features and I believe that it would be very helpful to be able to generate a seedlot from a trial. Then the plot weight recorded during harvest already being collected for the purpose of phenotyping would also be used as the initial inventory weight of the seedlot. I also noticed that 'seed multiplication' is one of the trial types, so it seems that a seedlot should always be connected to a source trial.

Possible implementation:

From the trial detail page, have a button that walks the user through creating new seedlots:

- Have option to create a Seedlot for every Plot or unique Accession
- Template for naming of Seedlots?
- optional: have user select a trait observed in the phenotyping trial that can be used to populate either the initial count or weight of the seedlot
- form for the rest of the seedlot metadata: location, box, etc (use breeding program of trial)

For Bugs:
---------

dwaring87:
Description 

This adds a workflow (from the Manage > Seedlots page) for creating Seedlots from the plots/accessions from a field trial.  It will create either one new seedlot per plot or per accession from the selected trial.  The user can also select a trait from the trial to be used as the value for each seedlot's initial contents.

Changes to the existing `seedlot-create` AJAX endpoint:
- Add a `plot_uniquename` parameter.  If this is provided, the plot will be used as the source of the first transaction created when creating the seedlot.
- Add a `no_refresh` parameter.  When this is set to 1, the stockprop materialized views are not automatically refreshed after creating the seedlot - they will have to be refreshed manually.

Fixes #3587 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/SGN/Controller/AJAX/Seedlot.pm
+++ b/lib/SGN/Controller/AJAX/Seedlot.pm
@@ -260,11 +260,12 @@ sub create_seedlot :Path('/ajax/breeders/seedlot-create/') :Args(0) {
     my $box_name = $c->req->param("seedlot_box_name");
     my $accession_uniquename = $c->req->param("seedlot_accession_uniquename");
     my $cross_uniquename = $c->req->param("seedlot_cross_uniquename");
-    my $source = $c->req->param("seedlot_source");
+    my $plot_uniquename = $c->req->param("seedlot_plot_uniquename");
     my $seedlot_quality = $c->req->param("seedlot_quality");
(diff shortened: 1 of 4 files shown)


lukasmueller (approved):
:-) :-)

PR #3604, opened 2021-07-02
Topic/solgs/outliers

isaak:
-- makes histogram plot interactive; mouseover on a bin shows a table of observations and their respective values; useful for identification of outliers/extreme performers.
-- adds phenotypic correlation analysis for data aggregated from multiple trials
-- refactored job submissions related code; created a controller for it;
-- refactored mason templates; removed duplication and consolidated overlapping templates into one
-- refactored phenotypic correlation R code
-- general code improvements and fixes
-- expand and pass selenium tests


lukasmueller (approved):
:-)

PR #3607, opened 2021-07-07
[BI-998] - Add additional info to brapi variables endpoint

ctucker3:
This is to facilitate adding tags to traits via the additional info field:

Something like this: 
```
{
  "observationVariableName": "yada yada",
  "additionalInfo": {
    "tags": ["favorites", "sea lice", "smolt"]
  }
}
```

Tested out:
- POST variables with and without additionalInfo. 
- PUT variables with additional info (overwrites existing), and without (deletes additional info)
- GET variables and POST search/variables both return additionalInfo

I'll wait to merge this until the work on bi-api and bi-web are done.


ctucker3:
Wrong repo, sorry

Issue #3534, opened 2021-05-20
Genotype download from protocol page has format problem


ClayBirkett:
When downloading genotype data for an accession on the protocol page the output does not have the GT data in the first field. This does not match the specifications for VCF file format and causes error in TASSEL program. The download works correctly in wizard.

lukasmueller:
Hi Clay, I was under the impression that this has been fixed...

ClayBirkett:
No, not fixed yet.

ClayBirkett:
fixed by pull request https://github.com/solgenomics/sgn/pull/3612

PR #3612, opened 2021-07-08
make GT field first in VCF as required by specifications

ClayBirkett:
in the genotype protocol page a download could not be opened in TASSEL program because the GT field was not the first field

#3534 

there is still a problem with sorting the VCF file but that is complicated by chromosome data that a non standard string instead of numeric

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


afpowell (approved):
Looks good to me

PR #3688, opened 2021-08-26
Topic/new plot name option on change accession

### LINKED ISSUE #3649: Updating plot accession creates confusion given plot name
Say you have an accession in a plot, MPA_173, the plot name is 21RYT_CNRRS-rep1-MPB_173_61 . When planting, you notice you don't have enough seed so you plant MPA_274 in that plot instead and update the accession for that plot to MPA_274. Now the accession in that plot is MPA_274, but the plot name is not updated in the system and  remains as 21RYT_CNRRS-rep1-MPB_173_61. The issue is Breedbase doesn't update the plot name when the accession is switched, and since the accession name is part of the plot name it leads to confusion. Either accession should not be part of the name (not required to make the plot name unique anyways) or the system should update the plot name when the accession is changed.

npanitz:
Description 
Resolves  #3649 by implementing new option to change plot name on replace accession features. Additional error handling added for new plot name and for existing plot name.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Trial/ParseUpload.pm
+++ b/lib/CXGN/Trial/ParseUpload.pm
@@ -17,7 +17,7 @@ has 'chado_schema' => (
 has 'trial_id' => (
   is => 'ro',
   isa => 'Str',
-  required => 1
+  required => 0
 );
 
 has 'filename' => (
--- a/lib/CXGN/Trial/ParseUpload.pm
+++ b/lib/CXGN/Trial/ParseUpload.pm
@@ -14,6 +14,12 @@ has 'chado_schema' => (
     required => 1
 );
 
+has 'trial_id' => (
+  is => 'ro',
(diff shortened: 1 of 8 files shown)


lib/CXGN/Trial/ParseUpload.pm
```diff
@@ -14,6 +14,12 @@ has 'chado_schema' => (
     required => 1
 );
 
+has 'trial_id' => (
+  is => 'ro',
+  isa => 'Str',
+  required => 1
```
nickmorales, line 20: This may cause problems with other functions

lukasmueller (approved):
:-)

PR #3692, opened 2021-08-27
materialized_markerview: cast pos to numeric

dwaring87:
Description 

This DB patch updates the definition of the `materialized_markerview` matview (as created by the `create_materialized_markerview` postgres function).  It changes the type of the pos column to numeric so that way comparisons are done numerically and do not need to be cast in the query.

The `cast(coalesce(nullif(s.value->>''pos'',''''),NULL) as numeric) AS pos` construct is used to handle instances where a marker's position is left blank/null.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
:-)

PR #3696, opened 2021-08-31
remove trailing slash at the end of hostname

isaak:
-- remove trailing slash at the end of hostname; doubles '/' when concatenating hostname with paths.
-- improve check for dataset categories


lukasmueller (approved):
:-)

PR #3727, opened 2021-09-20
solGS: Raw phenotype data model input viz

isaak:
-- add plot level phenotype data model input histogram visualization
-- add descriptive stat for both plot level and accession means level phenotype data
-- reorganize solGS analysis tools mason templates
-- add and pass selenium tests
-- refactor code


lukasmueller (approved):
:-)

PR #3731, opened 2021-09-22
identify username for account when resetting password

ClayBirkett:
We had several users who forgot their username so were not able to login or reset password
This change adds username to the email that gives link to password reset

After resetting password the screen was disabled with popup that could not be closed
This change adds redirect to home screen after succesfull password reset
goto home page after password reset

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
:-)

PR #3734, opened 2021-09-23
migrate tools to brapi v2

### LINKED ISSUE #3720: Brapi input cleaning bug breaks trial comparison and graphical filtering tool
Expected Behavior 

Clay and I did some investigating and the bug seems to come from the way the phenotypes-search call is handled by the Brapi controller.

The heatmap tool on the trial detail page issues the call like [this](https://github.com/solgenomics/sgn/blob/a05ac37d44313240db5b11b44569c6ad9b54aead/mason/breeders_toolbox/trial/phenotype_heatmap.mas#L444), then in the controller the params are processed by ```_clean_inputs()``` [here](https://github.com/solgenomics/sgn/blob/a05ac37d44313240db5b11b44569c6ad9b54aead/lib/SGN/Controller/AJAX/BrAPI.pm#L157) turning the ```$data_level``` scalar value into an arrayref. The Brapi CXGN modules expect an arrayref so it works fine

The trial comparison and graphical filtering tools use the phenotypes_search method from brapi.js, the params are stashed [here](https://github.com/solgenomics/sgn/blob/a05ac37d44313240db5b11b44569c6ad9b54aead/lib/SGN/Controller/AJAX/BrAPI.pm#L153) without being processed by ```_clean_inputs()```, so ```$data_level``` remains a scalar, leading to the error.

Not sure how to proceed from that. Using ```_clean_inputs()``` for all requests would fix it. But are POST requests for searches set up to not use clean inputs for a particular reason? Or would it be better just to skip ```_clean_inputs()``` in all cases and keep scalars as scalars? What's the point of the conversion?

For Bugs:
---------

```
error] Caught exception in SGN::Controller::AJAX::BrAPI->phenotypes_search
"Can't use string ("plot") as an ARRAY ref while "strict refs" in use at
/home/production/cxgn/sgn/bin/../lib/CXGN/BrAPI/v1/ObservationUnits.pm line 25, <$fh> line 1."
```

MFlores2021:
Description 
This fixes:
-Trial comparison 
-Graphical filtering

Close #3720

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/mason/search/stocks_graphicalfiltering.mas
+++ b/mason/search/stocks_graphicalfiltering.mas
@@ -57,12 +57,13 @@ function loadBrAPIData(server,parameters,success){
 
     var data = {
         "pageSize" : 10000000,
+        "includeObservations" : "true",
     };
     d3.entries(parameters).forEach(function(entry){
         data[entry.key] = data[entry.key]||entry.value;
     });
     
-    var brapi_Data = BrAPI("/brapi/v1",'v1.3',auth_token).phenotypes_search(data);
+    var brapi_Data = BrAPI("/brapi/v2",'v2.0',auth_token).search_observationunits(data);
 
     if (currentGFilter!=null){
         $("#stock_search_graphical_filter_results_div").DataTable().destroy();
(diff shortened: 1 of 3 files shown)


lukasmueller (approved):
:-)

PR #3785, opened 2021-10-30
fix job submitter email parsing

isaak:
fix job submitter email parsing and email analysis report to submitter


lukasmueller (approved):
:-)

PR #3837, opened 2021-11-22
preserve db stock names case-sensitive when comparing with names in file

isaak:
-- was first upper casing names before matching with names in file.


lukasmueller (approved):
:)

PR #3859, opened 2021-12-01
make download filename data type specific

isaak:
-- name download files after their respective data type (metadata.csv, phenotype.csv)


lukasmueller (approved):
:-)

PR #3869, opened 2021-12-07
remove extra column in table to fix DataTable warning

ClayBirkett:
Viewing the trial detail page causes a javascript error popup with DataTable error. This is caused by the genotyping protocols section that has an extra column. Removing the extra column "Sample Type" fixes the problem

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
:-)

review this to check whether sample type and create date should be there

PR #3958, opened 2022-02-18
Topic/solgs/pca add scree plot

isaak:
-- adds scree plot to PCA output
-- improves code
-- adapt and pass pca and clustering selenium test


lukasmueller (approved):
:-)

PR #3989, opened 2022-03-09
Topic/solgs/clustering evaluation metrics

isaak:
-- add Elbow plot to output
-- add and make cluster means, variance output downloadable
-- pass selenium test


lukasmueller (approved):
:-)

PR #3993, opened 2022-03-10
speed up genotyping protocols query

isaak:
-- populating genotyping protocols menu in the analyses pages that use genotype data was getting slower and slower (>30sec) as the genotyping_protocolsXtrials matview is getting larger. Querying the genotyping_protocols matview solves the problem. Hopefully, only genotyping_protocols with genotype data are in the production DBs.


lukasmueller (approved):
:-)

PR #3996, opened 2022-03-14
redirect to referrer page after login

isaak:
-- login was redirecting to homepage. Now, it redirects to the referrer page.


lukasmueller (approved):
:-)

PR #4011, opened 2022-03-18
Speed Improvements: Accession List Validation & Accession Properties Download

dwaring87:
Description 

This improves the speed of:

- The accession list validation:
    - For a list of 10,000 accessions it would take ~8 minutes to validate, now it takes ~2 seconds
    - The validation function now makes 1 query to find all of the items that have an exact (case-sensitive) match on the uniquename
    - Then it performs additional searches (case-insensitive uniquename and synonyms) for the items that did not have an exact match 

- The accession properties download (from `/breeders/download`):
    - For a list of 10,000 accessions it would take almost 30 minutes (validation ~8 minutes + download ~18 minutes), now it takes <10 seconds
    - The download function was creating a new Accession/Stock object for each list item, now it makes 1 query to get all of the properties for all of the list items

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/List/Validate/Plugin/Accessions.pm
+++ b/lib/CXGN/List/Validate/Plugin/Accessions.pm
@@ -27,71 +27,77 @@ sub validate {
     my @synonyms;
     my @multiple_synonyms;
     
-    foreach my $item (@$list) {
+    # First filter out exact matches
+    my $rs = $schema->resultset("Stock::Stock")->search({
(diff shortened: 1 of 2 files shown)


lukasmueller (approved):
:-)

PR #4024, opened 2022-03-28
Change Genotype Dosage

### LINKED ISSUE #3906: Allele dosages in VCFs count REF allele, _should_ count ALT to match header, marker-effects
I have noticed a possible bug with allele dosages on Cassavabase. 

I  am looking at data in the:
Gentoyping Protocol: IITA DArT-GBS 08 Aug 2021

Allele dosages in downloaded data appear to count the REF allele. 
According to the header of the VCF:
`##FORMAT=<ID=DS,Number=A,Type=Float,Description="estimated ALT dose [P(RA) + P(AA)]">`

However, DS field in the VCF, which I downloaded using the Wizard, as well as corresponding tab-separated dosage file both appear to count the REF allele. Can someone confirm this? I could be crazy :) ? Or it could only be this dataset.

I think the DS field is calculated by Cassavabase, is that correct? I guess o because the source VCF for the genotyping protocol (which I provided) does not have the DS field. I very much appreciate that the DB computes dosages and adds them to the VCF! Just need to tweak it.

**My use case / why this matters:**
I am doing predictions that use the _phased haplotypes_ extracted from the VCF file. I need the genomic-predicted marker-effects (computed using the dosages as predictors in an RR-BLUP model) to measure the effect of the "1" allele in the VCF (the ALT allele). To ensure this, in the past, I have used the haplotypes extracted from the VCF to manually compute my own dosages (summing the two haplotypes for each individual). 

Hope this makes sense, happy to clarify! Thanks!

nickmorales:
Description 
A db patch checks the saved genotyping protocols to see if the old dosage scheme was used, and then in a transaction updates the dosage for all stocks in those protocols.
On parsing of the genotyping data upload, the new dosage scheme is used.

closes #3906 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/db/00158/GenotypeDosageChange.pm
+++ b/db/00158/GenotypeDosageChange.pm
@@ -85,6 +85,8 @@ sub patch {
 	;";
     my $h1_1 = $schema->storage->dbh()->prepare($q1_1);
 
+    # GETS MARKER INFORMATION FOR ALL PROTOCOLS LOADED UNDER vcf_map_details
+
     my %protocol_geno_check;
     while (my ($nd_protocol_id, $markers_array_json) = $h->fetchrow_array()) {
         my $markers_array = $markers_array_json ? decode_json $markers_array_json : [];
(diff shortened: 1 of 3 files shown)


lukasmueller (approved):
:-)

PR #4043, opened 2022-04-11
fix f calculation for polyploids

labroo2:
my mistake, sorry!

Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/R/GCPC.R
+++ b/R/GCPC.R
@@ -199,8 +199,13 @@ write("GENETIC PREDICTIONS...", stderr())
 # in the matrix D before adding it to the estimated dominance marker effects
 # One unit of change in f represents changing all loci from homozygous to heterozygous
 
+if(userPloidy == 2){
 GC <- G - (userPloidy/2) #this centers G
 f <- rowSums(GC, na.rm = TRUE) / apply(GC, 1, function(x) sum(!is.na(x)))
+}
+if(userPloidy > 2){
+  GC <- G * (userPloidy - G) * (2 / userPloidy)^2
+}
 
 # Another alternate way to construct f is the total number of heterozygous loci in the individual
 # The coefficient of this construction of f does not need to be divided by the number of markers


R/GCPC.R
```diff
@@ -199,8 +199,13 @@ write("GENETIC PREDICTIONS...", stderr())
 # in the matrix D before adding it to the estimated dominance marker effects
 # One unit of change in f represents changing all loci from homozygous to heterozygous
 
+if(userPloidy == 2){
 GC <- G - (userPloidy/2) #this centers G
 f <- rowSums(GC, na.rm = TRUE) / apply(GC, 1, function(x) sum(!is.na(x)))
+}
+if(userPloidy > 2){
+  GC <- G * (userPloidy - G) * (2 / userPloidy)^2
```
lukasmueller, line 207: what are the two encodings that we are converting here... this is not clear to me :-O

R/GCPC.R
```diff
@@ -199,8 +199,13 @@ write("GENETIC PREDICTIONS...", stderr())
...
+}
```
lukasmueller, line 208: Shouldn't we also calculate an f value here, as in the diploid case?

lukasmueller (changes requested):

R/GCPC.R
labroo2: Yes! My mistake. 

if(userPloidy == 2){
GC <- G - (userPloidy/2) #this centers G
f <- rowSums(GC, na.rm = TRUE) / apply(GC, 1, function(x) sum(!is.na(x)))
}
if(userPloidy > 2){
  GC <- G * (userPloidy - G) * (2 / userPloidy)^2
  f <- rowSums(GC, na.rm = TRUE) / apply(GC, 1, function(x) sum(!is.na(x)))
}

labroo2: Lukas, sorry about this: there don't need to be if statements. This code works for any ploidy level:

  GC <- G * (userPloidy - G) * (2 / userPloidy)^2
  f <- rowSums(GC, na.rm = TRUE) / apply(GC, 1, function(x) sum(!is.na(x)))

lukasmueller: Oh, yes, that's super to do it without the ifs. I will add this to the pull request if you haven't done so yet. I
One strange thing is that I seem to get fractional values when I calculate the scores in my head for polyploids, is that really the desired outcome or is my math just wrong?

labroo2: Thanks so much! I did not add it to the pull request.
Yes, the fractional values are desired and correct. They are consistent with a digenic dominance model for autopolyploids. There are other models as well but I don't know much about them. I guess the idea is that the middle heterozygote is "most" heterozygous-- e.g. in an autotetraploid with 0 1 2 3 4, the 2 genotype-- so it gets the largest value of 1. The other two heterozygotes 1 and 3 are considered less heterozygous, since they have more similar allele copies overall, but equally so. So they get values of 0.75. 
Thanks again!

lukasmueller:
Added the code
GC <- G * (userPloidy - G) * (2 / userPloidy)^2
f <- rowSums(GC, na.rm = TRUE) / apply(GC, 1, function(x) sum(!is.na(x)))
to the topic/gpcp_tool branch directly. 

Closing this pull request.

PR #4084, opened 2022-05-11
Add option to rename_stocks.pl script to keep old name as a synonym

lukasmueller:
Description 
change the bin/rename_stocks.pl script such that the old name can be stored as a synonym using option -n (synonyms will be saved by default, -n will not save them).

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


ch728:
This looks like a useful change to the script. I am wondering if there could be a similar option on the front end. For example, when the user updates the unique_name on the web interface a pop-up could ask if they would like to add the previous name as a synonym.

MFlores2021 (approved):

PR #4154, opened 2022-07-05
Topic/solgs/download raw data

isaak:
-- make training population raw data and model input data downloadable
-- make model output data downloadable file
-- improve code


lukasmueller (approved):
:-)

PR #4174, opened 2022-07-16
Topic/solgs/download data 071522

isaak:
-- add training genotype and phenotype data and model output file downloads in  combined trials model pages


lukasmueller (approved):
:-)

PR #4205, opened 2022-08-25
Topic/solgs/change geno protocol issues

isaak:
-- improve use of alternative genotyping protocols in analysis tools
-- improve handling of out of range kinship values


lukasmueller (approved):
:-)

PR #4212, opened 2022-08-31
Topic/solgs/refactor 260822

isaak:
-- improve search for trials relevant for GS
-- add selection population genotype data download link
-- update and pass selenium test


lukasmueller (approved):
:-)

PR #4247, opened 2022-10-12
Topic/stat ontology dbpatch

isaak:
Description 
-- a dbpatch for loading statistics cvterms

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/db/00163/AddGenomicPredictionStatCvterms.pm
+++ b/db/00163/AddGenomicPredictionStatCvterms.pm
@@ -1,0 +1,72 @@
+#!/usr/bin/env perl
+
+=head1 NAME
+
+AddGenomicPredictionStatCvterms.pm
+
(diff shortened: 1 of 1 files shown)


nmenda (commented):
I think this should not go in a db patch. Maybe instead write a shell script in sgn-devtools for running the cvterms loader? Different servers can have the script in different directories too.

nmenda (approved):

nmenda:
since SGNSTAT needs to be updated in every instance of BreedBase a db patch is the right way to do this. Just make sure to use chado_tools and add a description of the SGNSTAT changes in the patch

lukasmueller (approved):
:-)

PR #4277, opened 2022-11-04
Topic/plot map fixes

dwaring87:
Description 

This makes a couple of small changes to the single and multi-trial uploads:

- It changes the plot row/col checks to warnings instead of errors (added in PR #4236)
- It changes the single-trial upload to handle warnings like the multi-trial upload
    - if warnings are encountered, the trial is not automatically saved
    - warning messages are displayed in the upload dialog
    - the user can override the warnings with an "ignore warnings" checkbox

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/js/source/legacy/CXGN/BreedersToolbox/UploadTrial.js
+++ b/js/source/legacy/CXGN/BreedersToolbox/UploadTrial.js
@@ -276,7 +276,11 @@ jQuery(document).ready(function ($) {
                 $("#upload_trial_error_display_second_try").show();
             }
             if (response.warnings) {
-                alert(response.warnings);
+                warnings = response.warnings;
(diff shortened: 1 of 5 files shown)


bellerbrock:
Hey @dwaring87, the updated warning reporting in the single trial upload looks great!

I'd worry that making duplicate row/col coords return a warning instead of an error could lead to problems with overlapping field trial plots, if/when users aren't careful about following up on warnings. Can we restrict that to the specific trial type where it's needed?

bellerbrock (approved):

darekbienkowski (approved):
:-)

PR #4327, opened 2022-12-06
refactor treatment query in wizard download

### LINKED ISSUE #4326: Some wizard phenotype downloads never complete in Cassavabase
Expected Behavior 

For Bugs:
---------

### Environment 

Cassavabase

#### Steps to Reproduce

Has been happening for about a month. Prasad sees the problem when trying to download data for these 4  accessions specifically : IITA-TMS-IBA070337, IITA-TMS-IBA011368, IITA-TMS-IBA011371, IITA-TMS-IBA120008

Looking at the error log and pg_stat_activity, it seems related to the new treatment retrieval query. It uses observation unit ids and placeholders, the queries remain active but don't take up many resources and never finish. May be exceeding the query length, a solution could be to modify the query to just use trial ids, then filter the results by the observation unit ids afterwards.

bellerbrock:
Description 

Shortens the treatment query lengths in wizard downloads by using trial ids then filtering the results by observation unit ids

closes #4326

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/BrAPI/v2/ObservationTables.pm
+++ b/lib/CXGN/BrAPI/v2/ObservationTables.pm
@@ -69,7 +69,7 @@ sub search {
     my @trait_names = @header_names[39 .. $#header_names];
     my @variables;
     foreach my $t (@trait_names) {
-        if ($t eq 'notes') { next; }
+        if ($t eq 'notes') { last; }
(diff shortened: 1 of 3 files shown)


lukasmueller (approved):
:-)

darekbienkowski (approved):
:-)

PR #4346, opened 2022-12-12
allow missing values to be specified with dot '.' for VCF input

### LINKED ISSUE #4328: Genotyping "shift" issue when using "." in vcf upload/download
Expected Behavior 

As reported by Jeffrey Endelman:

To clarify, the problem is not that we are missing markers. The markers are there, but Breedbase removed the two fields with “.” when exporting the data, shifting all of the data on that row. Here is a toy example:

 

INPUT file

#CHROM               POS         ID            REF         ALT         QUAL      FILTER    INFO       FORMAT                 Sample1

chr01                      100         m1          A             B              .               .               .               GT:AD                     0/1.:44,88

chr01                      110         m2          A             B              .               .               .               GT:AD                     1/1:0,81

.                               .               m3          A             B              .               .               .               GT:AD                     1/1:2,72

 

OUTPUT

#CHROM               POS         ID            REF         ALT         QUAL      FILTER    INFO       FORMAT                 Sample1

chr01                      100         m1          A             B              .               .               .               GT:AD                     0/1.:44,88

chr01                      110         m2          A             B              .               .               .               GT:AD                     1/1:0,81

m3                          A             B              .               .               .               GT:AD     1/1:2,72

For Bugs:
---------

ClayBirkett:
replace missing values in POS field of VCF output with 0

when values in VCF were specified as missing with dot '.' the transform_matrix.pl would remove those columns. This patch fixes that problem. The VCF specification also say that the POS column must be an integer and the TASSEL program will fail if POS is not an integer. To fix this case, the VCF output converts the dot '.' entry to a 0

fixes #4328

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/bin/transpose_matrix.pl
+++ b/bin/transpose_matrix.pl
@@ -2,6 +2,14 @@
 # 'transpose' swaps rows and columns in the given tab-delimited table.
 # syntax perl transpose.pl input.txt > output.txt
 
+use strict;
+use warnings;
+
+my @line;
+my @outline;
+my $lastcol;
+my $oldlastcol;
+
 while (<>) {
   chomp;
   @line = split /\t/;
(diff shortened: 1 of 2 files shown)


lukasmueller (approved):
:-)

afpowell (approved):

PR #4457, opened 2023-02-22
BrAPI: Search Observations Units - includeObservations param

dwaring87:
BrAPI Search Observations Units:

when the `includeObservations` parameter is included, it is being converted to an array somewhere, but the search function is expecting a "true"/"false" value.  This checks to see if the param is an array and uses the first value if it is.

This was causing the Compare Trials tool to fail since no observations were being returned from the BrAPI call.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


MFlores2021 (approved):
thanks!

darekbienkowski (approved):
:-)

PR #4488, opened 2023-03-14
Trait Validation: Check trait name for COMP traits

dwaring87:
Description 

### The Problem

When uploading trait observations, the column header uses the format of `trait_name|trait_id` (such as `Grain yield - kg/ha|CO_321:0001218`).  The trait validation just checks to make sure the DB and Accession (`CO_321:0001218`) are valid and ignores the trait name (this allows users to use different trait names, such as just `yield|CO_321:0001218`).

This isn't a problem for traits in the official trait ontology, since all of the traits share the same DB and Accession for all of the instances using the same crop trait ontology.

However, problems can arise for the composed traits.  If we add the same composed trait to two different databases, they will most likely have a different accession in each database (`COMP:0000043` = `Canopy Cover - UASHub - %|day 101` on DB 1 and `Canopy NDVI - UASHub - index|day 300` on DB 2).  If the user doesn't know the COMP Accessions are different in the different databases, the observation upload will be successful, but the data will be associated with a different trait.

### The Fix

This adds an additional check to the trait validation.  If the trait is in the COMP trait ontology, the validator will also check to make sure the trait name matches the name in the database.   All other traits still don't need to have their names match.

The error message displayed in the observation upload will include any matching traits (by trait name) that are potential replacements for missing traits (that don't have a valid ID for regular traits or matching trait name and ID for COMP traits).

image

### Configuration

By default, the trait validation will not include the additional name check.  To enable the name check for composable traits, set the `composable_validation_check_name` key to `1` in the `sgn_local.conf` file.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/BrAPI/v1/Observations.pm
+++ b/lib/CXGN/BrAPI/v1/Observations.pm
@@ -134,6 +134,7 @@ sub observations_store {
         has_timestamps=>1,
         metadata_hash=>\%phenotype_metadata,
         #image_zipfile_path=>$image_zip,
+        composable_validation_check_name=>$params->{composable_validation_check_name}
     );
 
     my ($stored_observation_error, $stored_observation_success, $stored_observation_details) = $store_observations->store();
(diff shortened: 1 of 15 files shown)


lukasmueller (approved):
:-)

nmenda (approved):

PR #4523, opened 2023-05-02
Topic/solgs/standalone correlation

isaak:
-- run correlation analysis from the analyze menu  using lists and datasets
-- refactor code
-- pass selenium tests


afpowell:
Hi Isaak - the correlation analysis looked fine to me, but I noticed two potentially related things:

1 - When I clicked on "Create a new list/dataset", the Wizard page that showed had the "Search Wizard" title doubled. This happened both on my local instance and on cassava devel

2 - I tried creating some new lists and datasets, but they did not show on the tool's dropdown. I was logged in and they showed up elsewhere, but not in the correlation tool. This also seemed to happen in both my local instance and cassava devel.

afpowell (approved):
:)

bellerbrock:
@isaak @afpowell I've tested a little more on my local machine. The standalone correlation tool's list input only shows me lists of type 'trials' and type 'plots', is that expected? The text 'Select a correlation population' makes me expect to see a dropdown of populations from the database, or a dropdown of lists with type 'accessions'

Also, I was wrong about where to find an example of adding a 'refresh' option to a list select, the wizard is not a good example. A better example is the second step in designing a new trial. In that step a refresh option is added to the dropdowns for selecting a list of accessions and checks. [Here's an example from the .js code](https://github.com/solgenomics/sgn/blob/3eb320fd18eebae7c74fe520ca2e4d06588db62b/js/source/legacy/CXGN/BreedersToolbox/AddTrial.js#L946), you just add 'refresh' as a fourth argument when calling the listSelect js function

Issue #3690, opened 2021-08-26
Heritability tool on Trial Detail page ( breeders/trial/###) pulls in data from other trials?


ch728:
Expected Behavior 
--------------------------------------------------------------------------
User expects that the heritability tool available on trial page will report heritability for that trial only.

For Bugs:
---------

Ve is reported in addition to Vg and Vresid. How can a Ve be reported for a single trial? The tool must be pulling data from other experiments, perhaps experiments from trials in the same folder that have overlapping accessions. I understand Breedbase keeps track of multi-environment trials by putting trials in the same folder at trial design, but still, doesn't make sense to have the tool on the Trial detail page pull this data in. Also, we have a folder ,  2021, that has singe location trials in it and they should not be assumed to all be apart of the same experiment by a tool.

### Environment 

Docker V29 Trial Detail page ( breeders/trial/###)

chris263:
This pull request  https://github.com/solgenomics/sgn/pull/4547 solves it

PR #4547, opened 2023-05-23
Topic/fixing names h2

chris263:
Description 

Fixing names in trials detail page.
Issue  https://github.com/solgenomics/sgn/issues/3690 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


js/.DS_Store
lukasmueller, line 1: I think this file should be removed from the git repo, as it is a Macintosh directory index file or something like that.

psql
lukasmueller, line 1: this file should probably also be removed

js/source/entries/heritability-old.js
lukasmueller, line 1: The new location for the js files should be in js/source/entries/ - so I wouldn't move a file from there to the legacy directory. Which should work on moving files the other way.

lukasmueller (commented):
A few minor comments in the code.

PR #4554, opened 2023-05-26
Checkbox to use tissue numbers 

Sri-2023:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Project.pm
+++ b/lib/CXGN/Project.pm
@@ -3757,7 +3757,7 @@ sub create_tissue_samples {
         if ($rs_previous_tissue->count > 0) {
             $previous_tissue_number = $rs_previous_tissue->first->value;
         }
-        my $new_tissue_number = $previous_tissue_number + scalar(@$tissue_names);
+        my $new_tissue_number = $previous_tissue_number + ($use_tissue_numbers ? scalar(@$tissue_names) : 0);
 
         my $treatments;
         my %treatment_experiments;
--- a/lib/CXGN/Project.pm
+++ b/lib/CXGN/Project.pm
@@ -3719,6 +3719,7 @@ sub create_tissue_samples {
     my $self = shift;
     my $tissue_names = shift;
     my $inherits_plot_treatments = shift;
+    my $use_tissue_numbers = shift;
     my $tissue_sample_owner = shift;
     my $username = shift;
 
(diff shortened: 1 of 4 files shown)


bellerbrock:
@Sri-2023 looks like the AddPlantsandTissueSamplesToTrial.t test is failing and needs to be updated, which makes sense since this PR modifies the `create_tissue_samples` subroutine. [Here's one place](https://github.com/solgenomics/sgn/blob/5548d23d7d40bc1b55d006debd2a1663b9c7fc15/t/unit_fixture/CXGN/Trial/AddPlantsandTissueSamplesToTrial.t#L1456) at least where the subroutine is used within the test and will need an extra value passed to it to represent the 'use_tissue_numbers' argument.

darekbienkowski (changes requested):
We need to describe functionality better.
At the moment, numbering works based on the number of tissues - as before - but now we also have tissues without numbers where they are still counted. That means if we add third tissue to a trial where the first two were without a number is still counted as third - like in a test that I updated. 
I don't know if it is correct behavior. 

```
  'test_trial25_plant_2_leaf',    # sample without tissue number
  'test_trial25_plant_2_root2',   # sample with tissue number
  'test_trial25_plant_2_fruit3'   # sample with tissue number

```

bellerbrock (commented):
The test fix looks good.

In terms of documenting or modifying the tissue numbering behavior - should it be specific to the tissue type/prefix? That would make the most sense to me. For example: 

  'test_trial25_plant_2_leaf',    # sample without tissue number
  'test_trial25_plant_2_root1',   # sample with tissue number
  'test_trial25_plant_2_fruit1'   # sample with tissue number
  'test_trial25_plant_2_fruit2'   # sample with tissue number

darekbienkowski:
Ok. Then I will later change the functionality not to count unnumbered tissues to a global tissues index.

darekbienkowski:
@Sri-2023 @bellerbrock I changed the test and Perl code for probably correct treating tissue without numbers, but can you check in UI if it works? I don't have a clue where that functionality is...

bellerbrock (approved):
The current version works fine for me. The checkbox makes it easy to turn off numbering, which is what Prasad and others were asking for.

And Darius's changes to improve the numbering when it is used looks good. Could enhance it even more to be prefix specific, but doubt that's worth the effort unless it's specifically requested.   

@darekbienkowski the code being changed can be tested in the web interface by going to Manage -> Tissue Samples -> then clicking on 'Create Tissue Samples for Field Trial'

darekbienkowski (approved):
;-)

PR #4618, opened 2023-08-12
KASP data upload (snp calls and raw data)

titima15:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (commented):
This feature should also be explained in the manual, what do you think?

titima15:
@lukasmueller can I add this with the intertek pull request?

isaak (approved):
Checking if an arrayref has a value [here](https://github.com/solgenomics/sgn/pull/4618/files#diff-08c9b870510f5ed295068ae328c50323d3fe5fdd5166ca2f66af4310ff7eafc8R110) (similarly in other places also) could be shorter and safer if you applied the same logic as in [here](https://github.com/solgenomics/sgn/pull/4618/files#diff-08c9b870510f5ed295068ae328c50323d3fe5fdd5166ca2f66af4310ff7eafc8R117).

titima15:
@isaak I need to do more testing about that modification. Seems like that method has been commonly used to check array ref for many objects. I'm still working on other part of genotyping upload, will look into it. Thanks

t/data/genotype_data/kasp_marker_info_with_facility_namesXXXX.csv
MFlores2021, line 1: is this test data needed? it seems similar to t/data/genotype_data/kasp_marker_info_with_facility_names.csv. Otherwise use a more meaningful name

titima15: No, It's not the same. This dataset will give warning.

titima15: I changed the file name to indicate file with error.

MFlores2021 (approved):

Issue #4611, opened 2023-08-04
Create a field view with multiple trials


chris263:
Expected Behavior 
--------------------------------------------------------------------------

Currently, it is only possible to view one trial per location. The idea here would be to provide the visualization of all trials in a specific area (farm). It would also allow the creation of "field management factors" for the area (this would include all trials in that area).

Then, it should generate view like that:

![Screenshot from 2023-08-04 12-29-02](https://github.com/solgenomics/sgn/assets/51970212/cd914543-ffd8-4f81-b576-934fe3b99b6b)

For Bugs:
---------

dwaring87:
This is something we've been wanting for T3 as well.  I worked on the plot map on the trial detail page so it can display plots from more than one trial *if the row and col positions are unique across trials* (ie trial 1 starts at row 1, trial 2 starts at row 10, etc).  Right now we don't have any way of linking the individual trials together into the same physical field, so it doesn't know which additional trials to pull from the DB.

chris263:
Maybe, if create an way to store geo coordinates for the 4 corners in each trial (In this case they don't need to store coordinates for all plots). Then, if possible adjust rows and columns inside those coordinates.
The second point is about management factors. If there is an way to add management factors per locations (including many trials) it will help.

MFlores2021:
fixed in https://github.com/solgenomics/sgn/pull/4620

PR #4620, opened 2023-08-16
Field Map: Display Multiple Trials

dwaring87:
Description 

This enables the display of trials that are located in the same physical field to be displayed on the same field map, if the trials match the criteria:

- They have the same year
- They have the same location
- The location type of the trials' location is "Field"
- The row and col positions of the plots for all of the trials don't overlap (ex: Trial 1 starts at row 1, Trial 2 starts at row 10, etc)

Our main use-case is that the planting order and harvest order downloads will include all of the plots from all of the trials in the same field since an entire field is often planted at once and not one trial at a time.

**Trial A Layout:**
Screenshot from 2023-08-16 08-30-38

**Trial B Layout:**
Screenshot from 2023-08-16 08-33-01

**Combined Layout:**
Screenshot from 2023-08-16 10-03-58

This should address part of Issue #4611

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


chris263 (approved):
Great!

MFlores2021 (approved):
:-)

PR #4623, opened 2023-08-18
Label Designer: support multiple trials, better sorting

### LINKED ISSUE #4604: Label Designer: Support more than one Trial
Expected Behavior 

We've received a number of requests for the label designer to be able to print plot-level labels for more than one trial at once.

The input source could be either a list of plots or a list of trials.

The labels should then be printed in a user-defined sort order:
  - trial name
  - plot number
  - row position
  - col position

For Bugs:
---------

dwaring87:
Description 

> [!IMPORTANT]
> **This depends on a [PR in the perl-local-lib repo](https://github.com/solgenomics/perl-local-lib/pull/3), which adds the `Sort::Naturally` module.**

This adds a couple of improvements to the label designer tool:

- Two-step source selection
  - I think some people weren't aware of the different data types (genotyping plates, lists, etc) that could be used as a source since they were added to the drop-down menu after all of the trials - and the list of trials can be very long
  - this adds a drop-down menu where you first select the data type of the source and then a second drop-down menu where you select the specific data item for the source (filtered by the selected type)
- Generate plot-level labels for more than one trial at once
  - added "plot details" as a label type if the source is a list of trials or a list of plots
- Sort labels by more than one property:
  - added the ability to sort the labels (in the generated PDF) by 1-3 properties (ex: first by trial, then by plot number)

----

**Source Selection:**

image

----

**Label Sorting:**
From the *Edit Additional Settings* button on the last step of the tool

image

Fixes #4604 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller:
The test might not pass because of the docker version in the github action. It should probably be version v0.80.

dwaring87:
@bellerbrock Good question.  It did seem like the labels were randomly sorted (and different each time) if there wasn't a sort option set.  I just added another commit that would sort by the item's database id if nothing was set.  This would at least generate the labels in the same order each time they're downloaded.  Since each source type will have different properties, I'm not sure what else would make a good default.

bellerbrock:
@dwaring yeah, the variable fields makes it tricky to set a default. Sorting by the db id / hash key is simple and at least makes it repeatable. But it still leaves the plots unordered within the trials I've looked at. I guess within-trial db ids are assigned via unordered hash rather then in order when a trial is created.

How about changing the design hash construction in `get_trial_design` to use names rather than db ids as the hash keys? That would then sort things better within trials by default. I can push a test commit with what I'm thinking of.

alockrow (approved):
Looks good! The [label designer manual](https://solgenomics.github.io/sgn/03_managing_breeding_data/03_12.html) will need to be updated too; should that have its own issue? It's also not obvious that you can only add multiple trials using lists: perhaps that could be shown in the updated manual as well.

PR #4633, opened 2023-08-31
Add ability to store additionalInfo for Lists

timparsons:
Description 
Add ability to store additionalInfo for Lists

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [x] New feature
  - [ ] User-Facing Change

### DIFF
--- a/db/00168/AddCvtermForListprop.pm
+++ b/db/00168/AddCvtermForListprop.pm
@@ -1,0 +1,100 @@
+#!/usr/bin/env perl
+
+
+=head1 NAME
+
+  AddListpropTable
+
+=head1 SYNOPSIS
+
+mx-run AddListpropTable [options] -H hostname -D dbname -u username [-F]
+
+this is a subclass of L<CXGN::Metadata::Dbpatch>
+see the perldoc of parent class for more details.
+
+=head1 DESCRIPTION
+
+This patch creates a listprop table to allow lists to store  additional info for brapi
+
+=head1 AUTHOR
+
+ Dave Phillips<drp227@cornell.edu>
+
+=head1 COPYRIGHT & LICENSE
+
+Copyright 2010 Boyce Thompson Institute for Plant Research
+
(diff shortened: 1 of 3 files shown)


MFlores2021 (approved):

db/00168/AddCvtermForListprop.pm
```diff
@@ -0,0 +1,100 @@
...
+    print STDOUT "\nChecking if this db_patch was executed before or if previous db_patches have been executed.\n";
+
+    print STDOUT "\nExecuting the SQL commands.\n";
+    my $schema = Bio::Chado::Schema->connect( sub { $self->dbh->clone } );
+
+    my $coderef = sub {
+        $self->dbh->do(<<EOSQL);
+
+ --do your SQL here
+INSERT into db (name)
```
nmenda, line 60: easier to do with the built in [create_with](https://metacpan.org/pod/Bio::Chado::Schema::Result::Cv::Cvterm#create_with) function 

We are now at patch folder 00177 and 00168 already exists

db/00168/AddCvtermForListprop.pm
```diff
@@ -0,0 +1,100 @@
...
+select 'null' where not exists(
+    select name from db where name = 'null'
```
nmenda, line 62: I would use a different db name, maybe 'local' ? 
Also need to update the system_cvterms file 
https://github.com/solgenomics/sgn/blob/master/system_cvterms.txt

nmenda (changes requested):

db/00168/AddCvtermForListprop.pm
timparsons: Updated

db/00168/AddCvtermForListprop.pm
timparsons: Both things have been completed

nmenda (approved):

Issue #4138, opened 2022-06-16
Phenotype Upload Fails with duplicate Stock names


dwaring87:
Expected Behavior 
--------------------------------------------------------------------------

We have a breeder who was able to upload a set of Seedlots and then create a Trial where the Plots in the Trial had the same set of names as the Seedlots.  So afterwards there as a Seedlot named "SI_Urb_22-1" and a Plot named "SI_Urb_22-1".

Then, she went to upload the phenotype observations for the trial, where the `observationunit_name` was set to the Plot name and the upload silently fails because the upload doesn't know which Stock to assign the values to.

The stock table has a constraint for unique organism/uniquename/type combination, so having the same name for different stock types is allowed at the database level.  Should the stock names be unique across all stock types?

The phenotype upload seems to lookup the stock entry just by name.  Should it filter the stocks by type (plot, subplot, plant, etc) so that way the Seedlot entry does not conflict with the Plot entry?

For Bugs:
---------

lukasmueller:
Code should check and not allow duplicate names even between different type_ids.

nmenda:
If we want to globally use only unique stock names then the stock object 'exists_in_database' function needs to be changed to lookup without type_id and organism_id 

https://github.com/solgenomics/sgn/blob/master/lib/CXGN/Stock.pm#L687

lukasmueller:
Hi Naama, I think that's the right thing to do... Would you like to implement this fix or should I go ahead and do it?

nmenda:
@lukasmueller I think I have a branch for this. Let me see if I can commit it

lukasmueller:
Is that function also used by the list checking functions? They may need to be changed as well...

nmenda:
@lukasmueller need to check everything that uses StockLookup 
https://github.com/solgenomics/sgn/blob/master/lib/CXGN/Stock/StockLookup.pm#L53https://github.com/solgenomics/sgn/blob/master/lib/CXGN/Stock/StockLookup.pm#L53

nmenda:
https://github.com/solgenomics/sgn/tree/topic/unique_seedlot_names 

Need to check other stock objects and make sure the uniqueness checks are not using type_id and/or organism_id

dwaring87:
@lukasmueller @nmenda This is still a problem.  It is still possible to upload a file of seedlots using plot names, which causes a lot of problems later on (when uploading trait observations with those plot names)

lukasmueller:
[...]

PR #4745, opened 2023-11-28
Seedlot Upload: check if name already exists as any stock

dwaring87:
Description 

This adds a check to the seedlot upload validation that returns an error if the seedlot name already exists as a stock name (excluding existing seedlots, which would be updated).

This addresses the immediate problem in Issue #4138

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


alockrow (approved):
Works if you Select "Add New Seedlot" though should it also check for duplicates when "Upload Seedlots" is selected?

Issue #4771, opened 2023-12-22
Pedigree Viewer: Scaled Too Small


dwaring87:
Expected Behavior 
--------------------------------------------------------------------------

We have some instances where the pedigree viewer is off-center and too small

![image](https://github.com/solgenomics/sgn/assets/7526014/5e5376bc-bcb1-4a6b-874c-9e015b141818)

![image](https://github.com/solgenomics/sgn/assets/7526014/a631f6ce-03fd-4843-b5d1-61aa232ab78a)

For Bugs:
---------

darekbienkowski:
https://github.com/solgenomics/sgn/pull/4798

PR #4798, opened 2024-01-27
fixed pedigree viewer when panel if collapsed

darekbienkowski:
Description 

{breedbase_address}/stock/XXXX/view

Changes in event trigger based on xpath. If something is wrong with the UI - it should trigger the pedigree viewer as before - without checking if the panel is open or not. 

image17

Screenshot from 2024-01-29 09-56-54

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


darekbienkowski:
https://github.com/solgenomics/sgn/issues/4771

MFlores2021 (approved):

nmenda (approved):

dwaring87:
@darekbienkowski I'm not sure if I'm having a problem with cached files or if there's something else going on, but I'm still having issues with the pedigree viewer.

This is the view when the panel is collapsed by default and then expanded:
image

This is the view when the panel is expanded by default:
image

Any thoughts on what to look into?

dwaring87:
I see a difference now:  I was collapsing the outer "Pedigree" section by default.  Collapsing the inner "Pedigree and Descendants" section by default works (when keeping the outer section expanded).

lukasmueller:
I think the inner section does not need a collapse capability

darekbienkowski:
> I see a difference now: I was collapsing the outer "Pedigree" section by default. Collapsing the inner "Pedigree and Descendants" section by default works (when keeping the outer section expanded).

Ok. I will check the behavior for the outer section.

dwaring87:
@darekbienkowski I just added a commit to this branch that moves the render trigger to the outer section instead of the inner section.  Can you take a look?

PR #4821, opened 2024-02-15
change matviews to native on brapi studies and observation units

### LINKED ISSUE #4652: Improve performance of uploading studies, observation units, and observations via BrAPI
Expected Behavior 
1. In lib/CXGN/BrAPI/v2/Studies.pm, update the store and update methods to:
    a. create the return list of studies in the code instead of by calling to the materialized view
    b. make the call to refresh the materialized view asynchronous
2. In lib/CXGN/BrAPI/v2/ObservationUnits.pm, update the observationunits_store and observationunits_update methods to:
    a. create the return list of observation units in the code instead of by calling to the materialized view
    b. make the call to refresh the materialized view asynchronous

Things to think about/investigate

- if materialized views are refreshed async, how does that affect when a user wants to download/view experiment and observations immediately after upload completes
- Can the use of the materialized views be removed from the BrAPI endpoints in BreedBase?
    - larger scope of change, but maybe more effective

_Notes from 9/7/23_

- BI to increase chunk size to 1000
- BB investigate changing BrAPI code to use the Native search instead of the Matviews search
    - removing the need to refresh matviews on save (line 838 BrAPI/v2/ObservationUnits.pm)
    - does there need to be a config to control if native or matview search is used?

_Notes from 9/14/23_

- @chris263 and @darekbienkowski to work on testing performance difference of fetching observations using matviews vs using native tables and share findings on 9/22/23
- @chris263 and @darekbienkowski to work on building a new query (or set of queries) to fetch observation and observation unit data in BrAPI form and share findings on 9/29/23

_Notes from 9/22/23_

> - @chris263 and @darekbienkowski to work on testing performance difference of fetching observations using matviews vs using native tables and share findings on 9/22/23

Fetching all observation units on the server -> matview returned in 7s, current native implementation returned in 270s
Fetching a single brapi study's OUs -> native was 25s

Fetching observations for a specific study -> matview returned in 2s, current native implementation returned in 40s

@darekbienkowski to evaluate current native query implementation to see if there are ways to trim it down

@timparsons and @nickpalladino to execute performance tests with chunk size of 100 vs 1000 by 9/29/23

Things to evaluate
- is there any code executing SQL queries within a loop? Can it be rewritten to execute one SQL query, and then process the results in the code loop?
- on create, are there any objects that don't have to query the database after they are created?
- evaluate performance of fetching observation units via BB UI vs BrAPI
- evaluate performance of fetching observation via BB UI vs BrAPI

For Bugs:
---------

MFlores2021:
Description 

-It creates a TrialLayoutSearch package to search by multiple trials using Native query. It returns observations when include_observations flag is pass (which will make it a bit slower ). 
-BrAPI calls ObservationUnits and ObservationTables were updated to use TrialLayoutSearch.
-BrAPI call Studies also were chaged from matviews to native.

Closes #4652

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


dwaring87 (approved):
I checked the field layout display for a few of our larger trials and the load time is a little bit longer, but not by much.

bellerbrock (approved):

PR #4822, opened 2024-02-15
In wizard downlod - require genotype protocol and warn if too many geno…

ClayBirkett:
If large genotype protocols are used the download will crash after a few minutes
These changes will warn user if number or markers * number of accessions is larger than one million
Also changes the wizard to require selection of genotype protocol

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/js/source/modules/wizard-downloads.js
+++ b/js/source/modules/wizard-downloads.js
@@ -54,13 +54,12 @@ export function WizardDownloads(main_id,wizard){
       selections["genotyping_protocols"]:
       [];
     main.select(".wizard-download-genotypes-info")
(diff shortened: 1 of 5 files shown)


ClayBirkett:
working on adding code so that 
1. if only project is selected then get related protocol, get number of markers, display protocol name, get chromosomes, update warnings
2. if no protocol or project then use default_genotype_protocol, get number of markers, display protocol name, get chromosomes, update warnings

lukasmueller (approved):
:-)

PR #4896, opened 2024-04-15
add markersets info popup and data validation

### LINKED ISSUE #4275: MarkerSets improvements
add documentation to manage markersets page, similar to "usage help" button on field layout. 
need to explain allele dosage and how missing data is handled
data type "download" - is this necessary?
data type "SNP" - would this be better as "Nucleotide"
add markers to markerset - could this be done as with a list, it is slow to add one by one
add markers to markerset - is there any validation for markers in markerset, it doesn't work when I validate using lists

ClayBirkett:
added "Usage Help" popup to manage markersets
added validation for marker type in "List"
commented out "Download" feature since it doesn't appear to be implemented

resolves #4275

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


titima15 (commented):
I'm wondering about the marker validation. With the current version which uses CXGN::Genotype::StoreVCFGenotype, markers are not stored in marker_alias table. 
 Which function stores markers in the marker_alias table?

ClayBirkett:
I believe that loading maps stores markers in the marker_alias table. In our database we have a materialized view of the markers but I don't believe that is in the general sgn database. I can try a query of the nd_protocolprop table but it may be slow.

titima15:
My main concern is that this plugin works only with specific set of data, not all markers. @lukasmueller and @nmenda might know more about markers stored in marker_alias table. Could you please advise on this?

ClayBirkett:
For marker type with json structure I switched to query nd_protocolprop table. The protocol_id is stored in the json so the query is quick. If the list is not json then there is no protocol_id and the validation will use the materialized_markerview

lukasmueller (approved):
:-)

titima15 (approved):

PR #4926, opened 2024-05-10
Phenotype upload: option to remove observations

### LINKED ISSUE #3017: Replace Phenotype Observations with NULL
Expected Behavior 

When uploading a phenotyping spreadsheet, if there is already a value stored for that trait and stock the verification step will give the user the option to overwrite the previously stored value with the new one.

image

However, is there a way to set an existing value to null or to remove the value?  When uploading a phenotyping spreadsheet with empty values for a trait, the verification screen does not prompt the user to overwrite the previously stored value and storing the file does not modify the values.

image

One use case for this scenario is that a breeder has identified outliers after the original data has been stored in the database and would like the outliers to be removed.

For Bugs:
---------

dwaring87:
Description 

Currently, when you upload phenotype observations, there is an option to replace any existing values with values from the upload file.  If the value in the file _is not_ blank, it will be used to replace the existing database value.  If the value in the file _is_ blank, the existing value in the database is left unchanged.

This PR adds an option (double-disabled by default), that allows the user to remove any existing observations from the database if the value is blank in the upload file.

This adds a `remove_values` attribute to the `CXGN::Phenotypes::StorePhenotypes` class (that is disabled by default).  Currently, the attribute is only used in the `StorePhenotypes` instance in the `/ajax/phenotype/upload_store` endpoint.

Fixes #3017 

image

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
:-)

MFlores2021 (approved):

PR #4962, opened 2024-06-03
Topic/doubly resolvable row column

chris263:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


bellerbrock (commented):
Nice work Chris! I tested with a typical 96 plot sweetpotato trial and the randomization in the resulting design looks good. 

I think rep should correspond to the additional blocking direction though, rather than having it the same as block, no? Check the layouts of DRRC trials that Russell has uploaded for examples. That way both blocking directions can be included in analysis models, and (once [this bug](https://github.com/solgenomics/sgn/issues/4967) is fixed), be seen in the fieldmap like this example: 

DRRCLayout

nmenda (approved):

chris263:
Hi Bryan, good point! I just fixed rep number. It might work now. Could you check again?
Thanks.

bellerbrock (approved):
Looks good!

Issue #4975, opened 2024-06-12
Generic Data File Parser


dwaring87:
Expected Behavior 
--------------------------------------------------------------------------

Create a generic data file parser that could be used with all of the file upload functions that will support .csv, .xls, .xlsx, .txt (tab-separated) files for any upload.  We've had a few people request CSV uploads due to problems/limitations with Excel.

I've started working on this on the [topic/generic_file_parser branch](https://github.com/solgenomics/sgn/tree/topic/generic_file_parser).

There is a new `CXGN::File::Parse` class that can be used to parse any of the supported file types into a uniform parsed data format.

For example:

```perl
my $parser = CXGN::File::Parse->new(
    file => '/home/production/public/data.csv',
    required_columns => [ 'accession_name', 'species_name' ]
);
my $parsed = $parser->parse();

my $errors = $parsed->{errors};
my $columns = $parsed->{columns};
my $data = $parsed->{data};
my $values = $parsed->{values};
```

will return:

- `errors`: an array of error messages encountered during file read / parsing
    - problems with opening the file (file doesn't exist, error from type-specific perl module)
    - missing required columns (when the required columns are specified)
    - rows with no values for required columns
- `columns`: an array of the column headers in the file
- `data`: an array of hashes, where each array item is one row of the input file
- `values`: a hash of the unique values for each column

Example Input:

| username | first_name | last_name     | email address       | organization       | address | country | phone | research_keywords | research_interests | webpage |
|----------|------------|---------------|---------------------|--------------------|---------|---------|-------|-------------------|--------------------|---------|
| testing1 | Test, Mȧ   |   Testing123  |  testing1@gmail.com | Cornell University |         |         |       |                   |                    |         |
|          |            |               |                     |                    |         |         |       |                   |                    |         |
|          |            |               |                     |                    |         |         |       |                   |                    |         |
| testing2 | Test       | Testing456    | testing2@gmail.com  | Cornell University |         |         |       |                   |                    |         |
| testing3 | John       | Testing       | testing3@gmail.com  | Cornell University |         |         |       |                   |                    |         |

Example Output:

```json
{
  "errors": [],
  "columns": [
    "username",
    "first_name",
    "last_name",
    "email address",
    "organization",
    "address",
    "country",
    "phone",
    "research_keywords",
    "research_interests",
    "webpage"
  ],
  "data": [
    {
      "address": null,
      "email address": "testing1@gmail.com",
      "first_name": "Test, Mȧ",
      "country": null,
[...]

PR #4983, opened 2024-06-24
Generic File Parser: Locations & Accessions

dwaring87:
Description 

This adds the `CXGN::File::Parse` class that can be used to parse any supported data file format into a uniform parsed data format (see #4975 for details).  It also updated the Locations and Accessions upload to use the new class.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lib/CXGN/File/Parse/Plugin/Excel.pm
```diff
@@ -0,0 +1,100 @@
...
+      $v = $super->clean_value($v, $hv);
+      $row_info{$hv} = $v;
+
+      if ( $v && $v ne '' ) {
+        if ( ref($v) eq 'ARRAY' ) {
+          if ( scalar(@$v) > 0 ) {
+            foreach (@$v) {
+              $values_map{$hv}->{$_} = 1;
+            }
+            $skip_row = 0;
```
titima15, line 78: Users have had issues when they manipulated excel files. Sometimes the files ended up having no data at the end of the files, but they are not actually empty. As a result, the loop is not ended. It helps after I changed from skip to last.

lib/CXGN/File/Parse.pm
```diff
@@ -0,0 +1,484 @@
...
+  my $data = $parsed->{data};
+  my $values = $parsed->{values};
+  my $required_columns = $parsed->{required_columns};
+  my $opional_columns = $parsed->{optional_columns};
+  my $additional_columns = $parsed->{additional_columns};
+
+COLUMN ALIASES
+  - Alternate column headers can be specified as column_aliases
+  - This will replace the alternate name with the preferred name in the final parsed data
+
```
titima15, line 88: Do we really need column aliases?

titima15:
I think it's very nice to have a generic file parser. The only draw back is that it aggregates data based on each row. Then we need to go over each row again to re-structure the data for validation step, depending on the type of data. It should be fine, just a bit slower, I guess.

titima15 (approved):
👍

dwaring87:
Changes from recent commits:
- stop parsing a file (Excel or Plain) if there are 5 blank rows in a row
- remove line on spreadsheet info about columns can be in any order (could be confusing to users)

I also tested a large accessions template (with 5000 entries).  The original parsing & verification step took 5.66 minutes, the generic parsing & verification took 5.69 minutes (so equally slow in both versions).

nmenda (approved):

PR #5017, opened 2024-07-16
Seedlots: Add support for subplot and plant sources

### LINKED ISSUE #4987: adding subplots as seedlots source
Expected Behavior 
When creating seedlots through the "upload seedlots" in the interface, there is the possibility to add a source, like a plot in a trial. But when using subplots, it doesn't recognize the source. Would it be possible to add the use of subplots as seedlots source please ?

For Bugs:
---------

### Environment 

Flaxbase, based on Breedbase v1.21

dwaring87:
Description 

This adds some changes to support subplot and plant entries as sources for seedlots:

- In the **Add New Seedlot** dialog: add optional source field that can be used to specify the name of a Seedlot, Plot, Subplot, or Plant that will be used as the source for the first seedlot transaction
- In the **Upload Seedlots** functions: allow the _source_ column to include plot, subplot, or plant entry names for both seedlots from named accessions or harvested from crosses
- In the **Create Seedlots from Trial** workflow: allow the user to generate seedlots from subplots or plants (if already added to the selected trial)

~~It also:~~
~~- fixes the **Create Seedlots from Trial** workflow to work with plots that use crosses instead of accessions~~
~~- adds an `accession_type` (the stock type of the plot's accession) to the trial layout design (to distinguish accessions from crosses)~~
    ~~- NOTE: ^ this will only be in the trial layout if it is freshly generated (not pulled from the cache stored in the projectprop table).  Should there be a DB patch that removes existing trial_layout_json values so new ones are generated?~~

Fixes #4987

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lib/CXGN/Trial/TrialLayout/AbstractLayout.pm
```diff
@@ -505,6 +505,7 @@ sub retrieve_plot_info {
 
     my $accession_name = $accession->uniquename;
     my $accession_id = $accession->stock_id;
+    my $accession_type = $accession->type->name;
```
titima15, line 508: For the source key in the layout, to prevent any downstream error, I used the original 'accession' key to indicate the source of the plot (accession, cross, family). Type of the plot source can be checked outside the layout.

titima15:
@dwaring87 If breeders would like to collect seeds from any plot/plant coming from a cross used in a trial, they would assign an accession name (as progeny from that cross) before collecting seeds. As a result, the source of that seedlot would be accession.

dwaring87:
The updates remove support from the "Create Seedlots from Trial" tool for trials that use crosses instead of accessions.  Now, when a trial is selected, it will check the stock type of the plots and display an error message if they're not all accessions.

alockrow (approved):

titima15:
@dwaring87 there is one important step missing during validation for creating a seedlot and uploading seedlots with source info. There is no check if the specified source (plot/subplot/plant) is derived from that accession. Would you like to add that check in this branch? I'm working on generic seedlot upload. If you prefer, I can also add that check in the branch I'm working on. Please let me know.

dwaring87:
@titima15 Thanks for noticing that - that is important.  I can add the check to this PR

titima15:
@dwaring87 one more thing, shouldn't we use "PlotsOrSubplotsOrPlants" for source validation?

dwaring87:
@titima15 The source could also be another seedlot or the contents accession / cross, right?

I added a contents check to the `create-seedlot` endpoint for adding a single seedlot.  It looks like the best place to add the check for the file upload would be the upload plugin's validation function.  Have you started working on the generic upload plugin?  If so, could you add the check to the new plugin?  If not, I can add it to the existing XLS plugin.

titima15:
@dwaring87 Yes, I'm working on the generic seedlot upload right now and I can add that check in the validate function.

titima15:
@dwaring87 If that seedlot has content as accession, it will validate if that name is an accession. Then the source can be on plot/subplot/plant. If that seedlot has content as cross, it will validate if that name is a cross. No source for cross seedlot.

titima15 (approved):
👍

titima15:
The validation works nicely. Thanks.

PR #5036, opened 2024-07-29
update solGS dataset and list pops selection interface

isaak:
Description 

-- update dataset/list selection interface for solGS

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (changes requested):
When I click on "Create training Population" a prompt is displayed if you would like to run it in the background, with the option "Yes" and "Cancel". Shouldn't there be an option "No" ?

lukasmueller (commented):
Maybe the Dataset set option should be the first one and the trials and "search a trait" interface should be below it?

afpowell (approved):
:+1:

lukasmueller (approved):
:-)

PR #5102, opened 2024-09-11
add season time events

### LINKED ISSUE #5095: additions to time ontology
Expected Behavior 
I proposed adding a 'categorical time of year' section of the cxgn time ontology following a similar format to the 'categorical time of day' section. This would address a few use cases - for example, a group typically collects 'early' 'mid' and 'late' season measurements, but the crop season extends from October to February. Of course, they could use the Julian week ontology terms, but would prefer not to as that information would be redundant with the measurements' time stamps. It would require them to create many additional composed traits and complicated downstream analysis- as the early, mid and late measurements do not consistently fall on the same week each year. 

As a rough start, I'd suggest the following:
Early Season - definition: first third of the growing season
Mid Season - definition: second third of the growing season
Late Season - definition: last third of the growing season. 

Additionally, season terms (Spring, Summer, Fall, Winter) would be helpful for similar reasons, but may also cause other complications.

Time ontology:
https://github.com/solgenomics/sgn/blob/master/ontology/cxgn_time.obo

nmenda:
Description 

closes #5095 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/ontology/cxgn_time.obo
+++ b/ontology/cxgn_time.obo
@@ -2909,3 +2909,21 @@ id: TIME:0000482
 name: post inoculation
 def: "After inoculation with a pathogen" []
 is_a: TIME:0000477 ! Event relations
+
+[Term]
+id: TIME:0000483
+name: early season
+def: "first third of the growing season" []
+is_a: TIME:0000477 ! Event relations
+
+[Term]
+id: TIME:0000484
+name: mid season
+def: "second third of the growing season" []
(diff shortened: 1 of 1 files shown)


isaak (approved):

lukasmueller (approved):
:-)

PR #5134, opened 2024-09-30
Seedlot Search: fix filter by count or weight

### LINKED ISSUE #5127: Seedlot search option minimum count not working
See below example. I put in 100, but seedlots with a count of 23 appear. 

image

dwaring87:
Description 

This fixes the Seedlot search when filtering by either count or weight by filtering out NAs and then casting the stockprop.value to numeric.

Fixes #5127 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


alockrow (approved):
Looks good! Interestingly seedlots with NA values will still come up if you set the minimums to 0 but that's probably not a problem as that doesn't seem like it would be an actual use case.

Issue #5111, opened 2024-09-17
VCF genotype upload problem


ClayBirkett:
The upload genotype VCF file is not working when selecting an existing protocol. The protocol_id is not set and the species is not retrieved from the protocol

https://github.com/TriticeaeToolbox/sgn/commits/sgn-396.0

#### Steps to Reproduce
Upload Genotype Data
Type of genotype data: VCF
select new project
select existing protocol

error message on web page say invalid null species
upload_genotyping_data_dialogs.mas

PR #5138, opened 2024-10-02
VCF upload allow new markers to existing protocol

ClayBirkett:
There is no code bug but there are lots of javascript warnings that led me to believe there was an error. My problem was when I tried to add a genotype project to existing protocol it failed because there was a new marker in the new project. This pull request will ignore those warnings when "Ignore any possible warnings" is selected during file upload.

I also remove the jquerymigrate module on production machines. This will eliminate many javascript warnings. The upload scripts use legacy javascript that have been deprecated for many years like js/source/legacy/jquery/iframe-post-form.js

#5111 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


ClayBirkett:
need to check if new markers are added to protocol
create configuration option to enable

PR #5150, opened 2024-10-14
Add extra checks for non-numeric values in weight and count fields during seedlot creation

### LINKED ISSUE #5149: Seedlot creation does not give error when user inputs string into seedlot count
Expected Behavior 
Seedlot creation should give an error if user types a string into seedlot count. Currently, it creates the seedlot and converts the string to an NA.

For Bugs:
---------

afpowell:
Description 
Gives user an error message if they enter non-numeric values in the seed amount or weight fields, rather than allowing the seedlot to be created.

Closes #5149

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/mason/breeders_toolbox/add_seedlot_dialogs.mas
+++ b/mason/breeders_toolbox/add_seedlot_dialogs.mas
@@ -207,7 +207,9 @@ jQuery(document).ready(function(){
         if (name == '') { alert("Please provide a name"); return; }
         else if (location == '') { alert("Please provide a location"); return; }
         else if (box_name == '') { alert("Please provide a box name"); return; }
-        else if (amount == '' && weight == '') { alert("Please provide an seed count amount or a weight in grams!"); return; }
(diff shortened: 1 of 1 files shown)


alockrow (approved):
Looks good. However, I noticed it won't accept a new seedlot with 0 count or 0 weight so there's now no option to make an empty seedlot. Is that a problem?

afpowell:
That's a great point, and I should definitely check with the group about it at the ticket meeting. The empty seedlot creation using 0s being disallowed is the current behaviour on master and, in terms of this PR, we still probably don't want to allow this weird 'backdoor' way of creating them by inserting character strings that then get turned into NAs, but it's definitely worth a discussion as to whether this should be enabled (via legitimate means)

isaak (approved):

PR #5159, opened 2024-10-21
Enable creation of empty seedlots

### LINKED ISSUE #5157: Enable creation of empty seedlots
Expected Behavior 
Currently, user can't create seedlot with 0 values for both seed amount and weight. User should be allowed to do this (i.e., create an empty seedlot).

For Bugs:
---------

afpowell:
Description 
This PR allows users to create empty seedlots as placeholders (i.e., seedlots with no/zero seed weight values or seed count values), as this seems desirable for certain users.

Closes #5157

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


alockrow:
Would it be better to require a weight or count but allow 0s? That way no one accidentally skips those fields without meaning to and there's still a value in one at least one of them.

Or, alternatively, you could give a warning message so that users have to confirm that they want to create a seedlot with no weight and no count.

ryan-preble:
> Would it be better to require a weight or count but allow 0s? That way no one accidentally skips those fields without meaning to and there's still a value in one at least one of them.
> 
> Or, alternatively, you could give a warning message so that users have to confirm that they want to create a seedlot with no weight and no count.

I'd say warning message makes the most sense to me. Without knowing who requested this feature I speculate that they want to be able to make a bunch of seedlot entries all at once in anticipation of seed collection and then go back and edit the data once they actually have the seeds collected. Giving a warning popup handles both this case and the case where someone just forgot to enter the data and needs to go back.

alockrow:
> > Would it be better to require a weight or count but allow 0s? That way no one accidentally skips those fields without meaning to and there's still a value in one at least one of them.
> > Or, alternatively, you could give a warning message so that users have to confirm that they want to create a seedlot with no weight and no count.
> 
> I'd say warning message makes the most sense to me. Without knowing who requested this feature I speculate that they want to be able to make a bunch of seedlot entries all at once in anticipation of seed collection and then go back and edit the data once they actually have the seeds collected. Giving a warning popup handles both this case and the case where someone just forgot to enter the data and needs to go back.

Sounds good to me!

lukasmueller (commented):
is the check that the value be numeric still there?

afpowell:
Yes, the check that a value be numeric should still be there

lukasmueller (approved):
:-)

lukasmueller (commented):
Actually, I found an issue: when I added 0 as the count, it appeared as NA in the seedlot later. Shouldn't it appear as 0? May be due to the fact that 0 is also false in Perl...

alockrow (changes requested):
Add a warning if the user chooses to make both count and weight empty to make sure that was on purpose.

alockrow:
There are a couple of issues I'm seeing:

1. If you just put in a space in the count and/or weight fields it will accept it and not throw an error and show up as an NA value in the seedlot. I think we should reject it and give the "a seedlot must have a count or weight" error.
2. As Lukas mentioned above if you put in a 0 it shows up as "NA". I think we want it to show up as a numeric 0.

afpowell:
Thanks, Amber - I'll check about the spaces and update this PR. 

On your other point, after some discussion, we are just going to use this PR to allow the user to create an empty seedlot by using zero values in either the amount or weight fields. Leaving both fields blank is not an option going forward. This will be followed by another ticket/PR that implements the user having commit

alockrow (approved):
Looks great!

PR #5160, opened 2024-10-21
Enable select all seedlots in available seedlots dialog

ryan-preble:
…toggled, all checkboxes in the open available seedlots modal are selected/deselected. List creation works. Disabled checkboxes are ignored.

Description 
Added a javascript function to enable users to select/deselect all available seedlots in the available seedlots modal. Disabled seedlots are ignored. Works with list creation. 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/mason/tools/available_seedlots.mas
+++ b/mason/tools/available_seedlots.mas
@@ -6,7 +6,7 @@
         <th rowspan="2" style="text-align: center;">Accessions</th><th colspan="7" style="text-align: center;">Seedlots</th>
       </tr>
       <tr>
-        <th style="text-align: center;"><input id="selectallaccessions" type="checkbox">Select All</input></th>
+        <th style="text-align: center;"><input id="selectallaccessions" type="checkbox" disabled>Select All</input></th>
         <th style="text-align: center;">Breeding Program</th>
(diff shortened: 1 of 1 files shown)


afpowell:
Hi Ryan, this looks really nice - I only had one minor comment: in the case where someone selects a list made up entirely of accessions that don't have any available seedlots, the Select All box is still enabled and clickable, which could be misleading. In that scenario, it might be good to have it disabled in the same way that the check boxes are for the individual accessions in the list.

ryan-preble:
> Hi Ryan, this looks really nice - I only had one minor comment: in the case where someone selects a list made up entirely of accessions that don't have any available seedlots, the Select All box is still enabled and clickable, which could be misleading. In that scenario, it might be good to have it disabled in the same way that the check boxes are for the individual accessions in the list.

Ok, I will change it to hide the select all button if the seedlots list is empty

afpowell (approved):
:+1:

Sri-2023 (approved):
👍

PR #5162, opened 2024-10-21
enable photo traits with brapi

lukasmueller:
Description 
Enable photo traits with brapi. Set trait_format to 'photo'.

Also, hide trait properties section for traits; only show for variables. Maybe this section should be renamed...

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/BrAPI/v2/Scales.pm
+++ b/lib/CXGN/BrAPI/v2/Scales.pm
@@ -432,15 +432,16 @@ sub _format_data_type {
     if ($value) {
         my %formats = (
 	    "categorical" => "Ordinal",
-            "numeric"  => "Numerical",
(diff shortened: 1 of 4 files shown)


bellerbrock (commented):
Tested with sptest and [this trait,](https://sweetpotatobase-test.sgn.cornell.edu/cvterm/76858/view/), it works!

It's outside the scope of this PR I think, but it'd be good to make the variable properties handling more sophisticated. Right now it is possible to add more than one trait format (though only the most recently added is shown in brapi). And the format is not restricted to subset of valid formats, but seems to accept any text input.

bellerbrock (approved):

PR #5177, opened 2024-10-31
display genotype data missing warning

isaak:
Description 
-- rename html elements attribute names for ease or understanding and maintainance
-- Show a message to a user to select a genotyping protocol  or upload genotype data before trying to use solGS tools that need genotype data,  in websites with no default genotyping protocol settings or no genotype data.
image

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/js/source/legacy/solGS/ajaxAutocomplete.js
+++ b/js/source/legacy/solGS/ajaxAutocomplete.js
@@ -20,7 +20,7 @@ jQuery(document).ready( function() {
 
 //population search autocomplete
 jQuery(document).ready( function() {
-        jQuery("#population_search_entry").autocomplete({
+        jQuery("#trial_search_box").autocomplete({
                 source: "/solgs/ajax/population/search",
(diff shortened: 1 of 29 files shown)


afpowell (approved):
:+1:
I would just recommend that, in the warning/error message text, you remove the "a" from the line "use the tool on a genotype data"

lukasmueller (commented):
On the fixture, there is some genotyping data and there is also a protocol. When I select the protocol, the message still appears. Shouldn't it disappear at this point?

isaak:
> On the fixture, there is some genotyping data and there is also a protocol. When I select the protocol, the message still appears. Shouldn't it disappear at this point?

Good catch. Fixed 481fd71. Many thanks.

lukasmueller (approved):
:-)

Issue #5113, opened 2024-09-19
Create an option to obsolete phenotype - QC


chris263:
Expected Behavior 
--------------------------------------------------------------------------
Currently, it is possible to obsolete an entire plot, which makes all phenotypes from that plot obsolete as well. The goal here is to allow obsoleting a single phenotype or trait within a plot while keeping the other values intact. This feature would be very useful for quality control on the trial details page.

For Bugs:
---------

lukasmueller:
There is an interface to mark a phenotype as a permanent outlier. Need to figure out the user interface to that.

PR #5191, opened 2024-11-07
Phenotype Quality Control

chris263:
This tool provides a Quality Control (QC) solution for phenotypic data validation. It allows users to visually inspect data through boxplots and detect outliers, which can be removed from downstream analyses. Only curators will have the privilege to restore raw data, ensuring data integrity is maintained while facilitating analysis.
Solve this tikect: https://github.com/solgenomics/sgn/issues/5113


js/source/entries/qualitycontrol.js
```diff
@@ -0,0 +1,493 @@
...
+    $(document).on('click', 'input[name=select_engine]', function (e) {
+        get_model_string();
+    });
+
+    $('#qc_analysis_prepare_button').click(function () {
+        dataset_id = get_dataset_id();
+        if (dataset_id != false) {
+            $.ajax({
+                url: '/ajax/qualitycontrol/prepare',
+                data: { 'dataset_id': get_dataset_id() },
```
isaak, line 25: use  dataset_id instead of get_dataset_id()

js/source/entries/qualitycontrol.js
```diff
@@ -0,0 +1,493 @@
...
+}
+
+
+
+
+function drawBoxplot(data, selected_trait, outlierMultiplier) {
+    const groupedData = d3.nest()
+        .key(d => d.locationDbId)
+        .entries(data);
+    if (outlierMultiplier === null){outlierMultiplier = 1.5}
```
isaak, line 358: What happens if the outlierMultiplier is undefined? May be checking for truthiness could be more general???

js/source/entries/qualitycontrol.js
chris263: Hey Isaak, It is running on cassava-test correctly. The get_data_id is working correctly there.

js/source/entries/qualitycontrol.js
chris263: There is a slider bar that changes the outlierMultiplier value, also there is a table showing minimum and maximum for the trait in each location. In this way it will allow users to define the cut off limit.

js/source/entries/qualitycontrol.js
isaak: Yes, I mean you already have stored the output from the get_dataset_id() method in the dataset_id. So, you could just use that...either way, it works...

isaak (approved):

afpowell (approved):

PR #5205, opened 2024-11-14
Generic Multi-Trial Upload

### LINKED ISSUE #5129: Multi-Trial Upload: fails with location codes
Expected Behavior 

[This earlier PR](https://github.com/solgenomics/sgn/pull/4476) added the ability to use the location abbreviation/code instead of the full name for the location in the multi-trial upload.

The upload now silently fails (says the trials were added successfully but are not added to the database) if one or more trials use the location abbreviation/code in the location column.  The upload plugin correctly validates the location code, but the additional location check in the `upload_multiple_trial_design.pl` script fails.

It might be a good idea to remove the location check in the perl script, since it's already being done by the plugin validation.

For Bugs:
---------

dwaring87:
Description 

- Add a new Multi Trial Upload Plugin using the generic file parser
- Deprecate the existing Excel Multi Trial Upload Plugin
- Replace the plugin used by the multi-trial upload workflow to just use the new plugin
    - before: it was reading the input file multiple times (once using the excel plugin in the AJAX controller, once using the generic file parser in the backend script, and again using the excel plugin in the backend script) 
    - now: the file is read just once using the new generic plugin 
- Move the file validation and parsing of the multi-trial template (by the plugin) from the AJAX controller to the backend script
    - since the trials are being saved in the backend script, the validation by the plugin should also be done in the backend script
    - the data being used to store the trials now uses the parsed data from the plugin (before: it was re-reading the original file and not using the parsed data from the plugin)
- Remove code to process phenotype observations from the backend script
    - this wasn't working since any additional columns (such as trait headers) are treated as management factors
- Consolidate where email messages are sent:
    - before: some emails were sent from the AJAX controller and others were sent from the backend script
    - now: all emails are sent from the backend script 
- This fixes problems causing the upload to fail if a location code was used (instead of the full location name) or the plot name is not given in the template
    - The location codes are replaced with the full location names in the upload plugin
    - Plot names are automatically generated in the upload plugin if not provided in the file
    - These issues would cause the upload to fail since the workflow was not using the parsed data from the plugin, now the trials are saved using the data from the plugin (with the location replacements made / plot names generated)
- Change how warnings are handled when storing in the background:
    - before: the trials would be stored if there were warnings encountered in the validation step (even if ignore warnings was not enabled before uploading)
    - now: trial storing will fail if warnings are encountered and the ignore warnings option is not enabled (the user will get an email stating that the trials were not stored due to the encountered warnings and that they should fix the warnings or upload the file again and enable the ignore warnings option) 

Fixes #5129 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


nmenda (approved):

Sri-2023 (approved):

Sri-2023:
Can we enable an auto-complete option for the email field that is associated with their account, if user opt to recieve results via email ? . And also, making it flexible for users to edit or change the email would be great !!!

Issue #5203, opened 2024-11-14
VCF upload skips the first 8 genotypes


lukasmueller:
Expected Behavior 
--------------------------------------------------------------------------

As reported by Laetitia Aznar

For Bugs:
---------

PR #5206, opened 2024-11-14
fix reading VCF with no header

ClayBirkett:
When the VCF has no header then the first few accessions are skipped while loading
This modification works on the transposed file

#5203 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller:
Instead of starting with the first line that contains number - slash - number, the parsing should simply start at the line after the line starting with #CHROM ?

lukasmueller:
In the code it actually skips 8 entries, which is correct for the untransposed file, but probably incorrect for the transposed file?

ClayBirkett:
I have it working now. The problem was that the parse_with_plugin() function was reading a fixed number of comment lines which is never true. I changed this to read all lines starting with "##", the comment lines. Then the next_genotype() function then skips the crhom, pos, id, ref, alt correctly

lukasmueller:
When I try to upload a test file, I see the following error in the log:
[error] Attribute (tempfile) does not pass the type constraint because: Validation failed for 'Str' with value undef at /home/production/cxgn/local-lib/lib/perl5/x86_64-linux-gnu-thread-multi/Moose/Object.pm line 24
	Moose::Object::new('CXGN::UploadFile', 'HASH(0x5adc3ba7e560)') called at /home/production/cxgn/sgn/bin/../lib/SGN/Controller/AJAX/GenotypesVCFUpload.pm line 380
	SGN::Controller::AJAX::GenotypesVCFUpload::upload_genotype_verify_POST('SGN::Controller::AJAX::GenotypesVCFUpload=HASH(0x5adc350a9e10)', 'SGN=HASH(0x5adc3ac95260)') called at /home/production/cxgn/local-lib/lib/perl5/Catalyst/Action.pm line 358
	Catalyst::Action::execute('Catalyst::Action=HASH(0x5adc354e9dd8)', 'SGN::Controller::AJAX::GenotypesVCFUpload=HASH(0x5adc350a9e10)', 'SGN=HASH(0x5adc3ac95260)') called at /home/production/cxgn/local-lib/lib/perl5/Catalyst.pm line 2060
	eval {...} at /home/production/cxgn/local-lib/lib/perl5/Catalyst.pm line 2060
......


and the following error on the user interface:

The organism species you provided is not in the database! Please contact us.

ClayBirkett:
I haven't seen that problem but for me the web VCF upload is still renaming my stock names and causing the upload to fail. The $include_lab_numbers setting is enabled when you choose "accessions" and that removes "." from stock names. I'll have to remove or fix that code.

alockrow:
@lukasmueller I was getting the same error until I made a new Genotyping Protocol instead of using the one already available on the fixture. It looks like the Protocol available on the fixture (GBS ApeKI genotyping v4) could be giving that error because its information format is out of date (see below).

image

lukasmueller:
We should update the fixture to have the correct format for the protocol... could that be done for in the context of this PR?

isaak:
From the cli:

Loading works, if there is at least one header line in the VCF file. 

If no headers, then the data about the protocol and markers is loaded, but it does not seem the alleles data is loaded. If you expand the 'Genotype Data' section, it hangs up  processing the retrieval of the data. No obvious error is thrown.

ClayBirkett:
If you are using the cassava_test.vcf, it may be the file is causing the loading problems. The last line is truncated. I cleaned up the file to make it valid.

lukasmueller (approved):

alockrow (approved):

Issue #5038, opened 2024-07-31
load cxgn statics obo in cassavabase and test


isaak:
Expected Behavior 
--------------------------------------------------------------------------

-- run in dockerized sites: 

perl /home/production/cxgn/**chado_tools**/chado/bin/gmod_load_cvterms.pl -H dbhost_db -D dbname -s SGNSTAT -r postgres -d Pg -p dbpass  /home/production/cxgn/sgn/ontology/cxgn_statistics.obo

--for sites still in virtual machines:
 perl /home/production/cxgn/**Chado**/chado/bin/gmod_load_cvterms.pl -H dbhost_db -D dbname -s SGNSTAT -r postgres -d Pg -p dbpass  /home/production/cxgn/sgn/ontology/cxgn_statistics.obo

For Bugs:
---------

isaak:
Done.

PR #5265, opened 2025-01-03
store pca scores results

isaak:
Description 
-- adds an option to store the score of  top 10 principal components. For now, the option is available if a user submitted the pca job analysis to run in the background and expected an email notification about the job status.

-- adds pca stat terms to the cxgn_statistics.obo file
-- updates pca selenium tests
-- updates fixture with the db patches and loaded cxgn_statistics.obo

Required:
-- run db patch
-- reload cxgn_ontology.obo
-- potentially delete double 'VARIABLE_OF' cvterm entry in the database

image

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


chris263 (approved):
:-)

ryan-preble:
I ran the db patches and tried to save PCA results and was met with an error... I sent a slack message about it

isaak:
> I ran the db patches and tried to save PCA results and was met with an error... I sent a slack message about it

Please run do this: https://github.com/solgenomics/sgn/issues/5038

and follow this message from @naama: 
log into your postgres db and search for 'variable_of' cvterm
[4:26](https://muellerlab.slack.com/archives/C04GGDDD7/p1732634787332219)
you will probably find 2, and you need to have only one with a cv_id of the 'relationship ontology'
4:28
you want to delete the duplicated 'variable_of 'cvterm , but before deleting it you want to update your cvterm_relationship table and do an update  begin; update cvterm_relationship set type_id = CORRECT_VARIABLE_OF_CVTERM_ID where type_id = CVTERM_ID_TO_BE_DELETED (edited)

ryan-preble:
> > I ran the db patches and tried to save PCA results and was met with an error... I sent a slack message about it
> 
> Please run do this: #5038
> 
> and follow this message from @naama: log into your postgres db and search for 'variable_of' cvterm
> 
> you will probably find 2, and you need to have only one with a cv_id of the 'relationship ontology'
> 
> you want to delete the duplicated 'variable_of 'cvterm , but before deleting it you want to update your cvterm_relationship table and do an update begin; update cvterm_relationship set type_id = CORRECT_VARIABLE_OF_CVTERM_ID where type_id = CVTERM_ID_TO_BE_DELETED (edited)

Tried this, still does not work. Current issues I get (running breedbase docker):
- Analysis details page is broken: 
image
- After running db patches and reloading ontologies `perl /home/production/cxgn/chado_tools/chado/bin/gmod_load_cvterms.pl -H dbhost_db -D dbname -s SGNSTAT -r postgres -d Pg -p dbpass /home/production/cxgn/sgn/ontology/cxgn_statistics.obo -u` The error persists. variable_of cvterm is not duplicated but is missing outright. Note that the test fixture does not appear to have this by default. 
- Saving PCA results is bugged. Trying to do two PCAs on one dataset results in the following output: 
image
This is a bit confusing because one cannot tell which PCA each save button belongs to. 
- Pressing the save button always results in error on first press saying each plot in the dataset (outliers_test_dataset from the test fixture in this case) is invalid. Error message fades away, and pressing the button a second time appears to allow one to view the saved PCs. But, following that link then takes you to the analysis details page, which is still broken. 

I suspect all of these issues are caused by the cvterm problem, but I do not know how to fix it

isaak:
The broken analysis detail page is fixed #5264. I merge master (with the fix) to the branch, so that should be resolved.

Could you reproduce how the results links are all showing at the bottom, instead of right below the corresponding plots?

It should be as below:
image

ryan-preble:
> The broken analysis detail page is fixed #5264. I merge master (with the fix) to the branch, so that should be resolved.
> 
> Could you reproduce how the results links are all showing at the bottom, instead of right below the corresponding plots?
> 

1. Use a dataset to create a PCA. Use the link in the user's home account to view that analysis on the PCA page. 
2. At the PCA page, run the exact same PCA using the same dataset. The results will be immediately loaded, but with two identical plots and two identical buttons.

isaak:
> > The broken analysis detail page is fixed #5264. I merge master (with the fix) to the branch, so that should be resolved.
> > Could you reproduce how the results links are all showing at the bottom, instead of right below the corresponding plots?
> 
>     1. Use a dataset to create a PCA. Use the link in the user's home account to view that analysis on the PCA page.
> 
>     2. At the PCA page, run the exact same PCA using the same dataset. The results will be immediately loaded, but with two identical plots and two identical buttons.

Thanks, Ryan. I reproduced it. If you followed the results link from your profile page, it should have loaded the results in the pca page like below. Doesn't it do the same for you? It is not necessary to click the run pca button again for the same dataset, otherwise the same output is displayed again. I will think about a solution for it anyway.

image

ryan-preble:
> > > The broken analysis detail page is fixed #5264. I merge master (with the fix) to the branch, so that should be resolved.
> > > Could you reproduce how the results links are all showing at the bottom, instead of right below the corresponding plots?
> > 
> > 
> > ```
> > 1. Use a dataset to create a PCA. Use the link in the user's home account to view that analysis on the PCA page.
> > 
> > 2. At the PCA page, run the exact same PCA using the same dataset. The results will be immediately loaded, but with two identical plots and two identical buttons.
> > ```
> 
> Thanks, Ryan. I reproduced it. If you followed the results link from your profile page, it should have loaded the results in the pca page like below. Doesn't it do the same for you? It is not necessary to click the run pca button again for the same dataset, otherwise the same output is displayed again. I will think about a solution for it anyway.
> 

Yes it works so long as one does not try to run the pca again for the same dataset. However, I am still getting an error with actually saving the results. When clicking the save PCs button, I get the following error message:
image
image
I can then click the save button again, and it will allow me to visit the analysis details page. This mostly works, however attempting to open the analysis results tab crashes it. 
image

Any idea why the plots would be invalid? It could be related to the data in the test fixture - not sure.

isaak:
> 2\. At the PCA page, run the exact same PCA using the same dataset. The results will be immediately loaded, but with two identical plots and two identical buttons.

fixed 7e74097

PR #5304, opened 2025-02-05
BrAPI: set default germplasm acquisitionDate and use in filtering

### LINKED ISSUE #5303: Get Cassavabase to work with the BrAPI Germplasm Search Tool
We'd like to get cassavabase to work with our BrAPI-enabled germplasm search tool: https://synonyms.triticeaetoolbox.org

However, due to the very large number of clones in cassavabase, the tool is unable to reliably fetch all of the germplasm records via BrAPI.

One workaround for this would be to allow the BrAPI call to only fetch germplasm entries from the last few years to limit the size of the query.  This could be done by filtering on the germplasm `acquisitionDate` property.

dwaring87:
Description 

This adds more support for the `acquisitionDate` germplasm property in BrAPI endpoints:
- The germplasm `acquisitionDate` property will use the `acquisition date` stock property if it was manually added for a particular accession.  When this is NOT present, it will use the stock `create_date` field (automatically set when a stock is added to the database) for the `acquisitionDate`.  This ensures that every stock will have an `acquisitionDate` set (either the `create_date` by default or the `acquisition date` stock property, if set).
- Add additional filter parameters to the `GET` `/germplasm` and `POST` `/search/germplasm` BrAPI endpoints:
    - `acquisitionDate` - filter germplasm for an exact acquisition date
    - `minAcquisitionDate` - filter germplasm that have an acquisition date >= the param value
    - `maxAcquisitionDate` - filter germplasm that have an acquisition date <= the param value 

Fixes #5303 

Related to plantbreeding/BrAPI#624

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
it's not clear how the search works with acquisition_date, as it is just a match and not a smaller or larger than comparison

dwaring87:
> it's not clear how the search works with acquisition_date, as it is just a match and not a smaller or larger than comparison

@lukasmueller  `acquisitionDate` is an exact match, `minAcquisitionDate` is a >= match and `maxAcquisitionDate` is a <= match.

Issue #5294, opened 2025-02-03
Infinite mystery rows parsed in Excel parsing


lukasmueller:
This has been observed for the accession upload but may also be happening in other uploads: Sometimes the Excel parser cannot determine the end of the "data zone" in the Excel file and continues parsing to the end of the Excel file, parsing 100,000s of rows and of course timing out in the process. We need to manually check the end of the parsing, such as checking for empty values.

It is not clear what causes this to happen, I think it is when columns or copied around potentially it will set the "data zone" to the entire column length.

afpowell:
@lukasmueller - Do you have a file that causes this?

dwaring87:
I did check the example file from Lukas and it does have a lot of "blank" rows (rows with no data, but are still included in the Excel file, likely because they have formatting information).  I don't think it is the file parser that is causing the slow upload - it will stop parsing the file after 5 empty rows.  The slow part seems to be the Fuzzy Search when uploading accessions.

dwaring87:
I'm closing this because I think enough checks have been added to the file parser to stop this particular use-case

PR #5321, opened 2025-02-12
Generic File Parser: Improve empty column handling

dwaring87:
Description 

This updates the Generic File Parser and the Excel and Plain text plugins to improve handling of empty columns.

If a column does not have a header, it will not process that column when processing each of the rows.
It will stop checking columns after 5 blank columns.

Related to #5294

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
:-)

PR #5349, opened 2025-03-04
Add seedlot material type

titima15:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/db/00191/AddMaterialTypeCvterm.pm
+++ b/db/00191/AddMaterialTypeCvterm.pm
@@ -1,0 +1,74 @@
+#!/usr/bin/env perl
+
+
+=head1 NAME
+
+AddMaterialTypeCvterm
+
+=head1 SYNOPSIS
+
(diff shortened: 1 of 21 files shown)


lukasmueller (approved):
:-)

dwaring87 (approved):

dwaring87:
@titima15 This works great!  A couple of suggestions (maybe for a future PR):

- The Material Type input on the main seedlot page (when searching) could be a dropdown menu with the types already set
- Would it be useful to have a default material type set in the sgn_local.conf file?  This could be the default for new seedlots - and maybe the default for older seedlots where the type is not defined.

titima15:
@dwaring87 Yes, Lukas and I were discussing about an option to set a default material type in the sgn_local.conf file. I will work on that. It's a good idea to add dropdown menu for searching, will do that. Thanks David.

PR #5383, opened 2025-03-24
Search Wizard: multi-line filter

### LINKED ISSUE #5375: Search Wizard: Multi-item filter
Currently in the Search Wizard, a user can filter a column of items by typing in the name of the item they're interested in into the Search field at the top of the column.  If they're interested in multiple items, they would have to find those individually in the column or create a list of the items first.

We've had requests from users who would like to be able to type in or paste in multiple items at once to use as a filter on the column.

dwaring87:
Description 

This expands the search filter in each column of the Search Wizard from a single-line text input to a multi-line textarea, where matches are found for each line of input - useful for filtering by multiple items or pasting in a list of items.

It also adds a "Contains" vs "Exact" match type.  Contains (the default and original behavior) will find matches the contain the case-insensitive search term in the database term and Exact will find an exact case-insensitive match.  For example, if the search term is "BP25-123", a "contains" match will match to "BP25-123" and "BP25-123-A" while an "exact" match will only match to "BP25-123".

Here's a demo:

https://github.com/user-attachments/assets/9bca6270-07b0-4e51-a9d8-1f2ec3afc813

Fixes #5375 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/js/source/modules/wizard-search.js
+++ b/js/source/modules/wizard-search.js
@@ -258,6 +258,17 @@ export function Wizard(main_id,col_number){
     reflow(d.index);
   }).filter(d=>d.index>0).select(".wizard-lists-group").remove();
 
+  allCols.select('.wizard-search-options-contains').on("click", function(d) {
+    toggleSearchOptionsMatch(this, d, "contains");
+  });
+  allCols.select('.wizard-search-options-exact').on("click", function(d) {
+    toggleSearchOptionsMatch(this, d, "exact");
+  });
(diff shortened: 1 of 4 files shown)


ryan-preble (approved):

lukasmueller (approved):
Amazing! 
:-)

PR #5384, opened 2025-03-25
Topic/analysis result stock type

isaak:
Description 
-- adds analysis_result stock type to cvterms
-- adds option to store analysis result as analysis_result stock type, as opposed to as accession (default).

Required:
-- dbpatch 00190/AddAnalysisResultStockType.pm

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


chris263:
Currently mixed models a and other loaded analysis are using 'analysis_instance'. Could other analysis also use 'analysis_instance'?
https://github.com/solgenomics/sgn/blob/master/db/00119/AddAnalysisAndMixedModelCvterms.pm

titima15:
I have the same question as Chris. Why we need additional stock type for storing analysis result? We already have 'analysis_instance" stock type (as well as other related cvterms) for storing results. What is the difference between analysis_result and analysis_instance? We can use stockprop to specify different types of analysis.  It seems like sometimes the 2 stock types are not clearly defined. For example, you have a new sub 'get_analysis_instance_stock_type', but actually the function retrieves analysis_result stock name.

lukasmueller:
> Currently mixed models and other loaded analysis are using 'analysis_instance', maybe there is no need to create 'analysis_result'. Sould all analysis use 'analysis_instance'? https://github.com/solgenomics/sgn/blob/master/db/00119/AddAnalysisAndMixedModelCvterms.pm

Hi Chris, I think that "analysis_instance" replaces what the "plot" is in a trial, but "analysis_result" replaces what the accession represents and as such does not yet exist.

chris263 (approved):

ryan-preble (approved):

PR #5439, opened 2025-04-23
Set brapi_require_login to 1 by default

afpowell:
Description 

This PR sets the default value for brapi_require_login to 1 in the sgn.conf file. It seems desirable to have this as the default now, since we are setting up more breedbase instances with data that needs to be private/protected.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/sgn.conf
+++ b/sgn.conf
@@ -68,7 +68,7 @@ require_login 0
 #### Report parameters
 report_engine phenotype_properties_check
 
-brapi_require_login 0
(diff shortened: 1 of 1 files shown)


lukasmueller (approved):
:-)

bm743 (approved):

PR #5457, opened 2025-05-06
Allow users to set experimental control for transformation

titima15:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


afpowell:
Some questions/comments about a few (very minor) things I noticed:
-If I set a transformation ID to 'is a control', the "Set Transformation Control" is still enabled... it seems like its two options ('Select a Control for this Transformation ID' and 'Set this Transformation ID as a Control') are then less relevant at that point? Should it be disabled or maybe just hidden?
-The 'Spreadsheet Format' info link isn't working for me on the workflow for 'Upload Transformation IDs' - looks like it may be missing that modal?
-When I click 'Add New Transformation ID', then check 'is a control', and add the ID, if I then try to click 'Add New Transformation ID' again, the 'is a control' box is still checked. We may just want this to be cleared/reset?

lukasmueller (approved):
:-)

afpowell (approved):
:+1:

Issue #3158, opened 2020-11-12
Implement field spatial correction algorithms


lukasmueller:
Expected Behavior 
--------------------------------------------------------------------------
Field irregularities can be compensated computationally, and corrected fields could be stored as analyses. What are the algorithms we should consider for this?
For example:
https://www.biorxiv.org/content/10.1101/612036v1.full.pdf

For Bugs:
---------

bauchetg:
INLA or SPATS https://cran.r-project.org/web/packages/SpATS/index.html have 2D splines corrections

chris263:
Added in mixed models

https://github.com/solgenomics/sgn/pull/5400

PR #5478, opened 2025-05-08
Topic/adding spatial correction

ryan-preble:
Description 
Supports the spatial corrections added to mixed models by @chris263 . In the trial detail page, under the phenotype heatmap, users can now calculate plot-level trait value adjustments based on the effects of spatial layout. Modeling done in SpATS in R.  Users can store the results both in association with the trial as a JSON of trait value adjustments and as a standalone analysis as a set of plots with corrected values. 

Requires merging R_libs changes made to topic/debian_bullseye_R4.4

#3158 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


R/Spatial_Correction.R
```diff
@@ -0,0 +1,75 @@
...
+    rowNumber = userPheno$rowNumber,
+    colNumber = userPheno$colNumber,
+    replicate = userPheno$replicate,
+    blockNumber = userPheno$blockNumber,
+    plotNumber = userPheno$plotNumber
+)
+
+for (trait in userResponse) {
+    spatial_model <- SpATS(
+        response = trait,
```
cmn92, line 62: should we have an option where the row and column are modelled as separate random factors besides as splines? 
random = ~ R + C whereby R = as.factor(rowNumber) and C = as.factor(colNumber)?

cmn92 (approved):
should we have an option where the row and column are modelled as separate random factors besides as splines? 
random = ~ R + C whereby R = as.factor(rowNumber) and C = as.factor(colNumber)?

R/Spatial_Correction.R
ryan-preble: This is a good question, and I think I need to ask Chris how spatial correction in mixed models is done. Since he uses rows and columns as random effects in mixed models, doing so here might lead to overcorrection or oversmoothing. But, this does get me thinking that I should probably include another moran test in the spatial corrections script to see whether spatial autocorrelation was really decreased.

isaak (approved):

PR #5489, opened 2025-05-14
create 'composed traits' cvterm if does not exist

nmenda:
Description 
error in okra.breedbase 
No cvterm for composed traits, so has to be created from the post composing tool the first time a COMP trait is generated 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Onto.pm
+++ b/lib/CXGN/Onto.pm
@@ -152,38 +152,32 @@ sub store_composed_term {
         $h->execute();
         my $accession = $h->fetchrow_array();
 
-      my $new_term_dbxref =  $schema->resultset("General::Dbxref")->create( {
-          db_id     => $db->get_column('db_id'),
-          accession => sprintf("%07d",$accession)
-      });
-
-      my $parent_term= $schema->resultset("Cv::Cvterm")->create_with(
-        { cv     =>$cv,
-          name   => 'Composed traits',
-      });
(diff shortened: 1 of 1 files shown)


lib/CXGN/Onto.pm
```diff
@@ -157,21 +157,21 @@ sub store_composed_term {
           accession => sprintf("%07d",$accession)
       });
 
-      my $parent_term= $schema->resultset("Cv::Cvterm")->find(
-        { cv_id  =>$cv->cv_id(),
+      my $parent_term= $schema->resultset("Cv::Cvterm")->create_with(
```
lukasmueller, line 160: shouldn't this be something like find_or_create_with? You want to create this only once?

lukasmueller (commented):
will this create duplicate entries of the composed variable?

nmenda:
@chris263 

> @nmenda , this is working fine to add new post composed traits. However, it is showing post comp as "null" in the search page.
> 
> Screenshot from 2025-05-14 20-16-15

this was a database issue with the db_id of the root term for the COMP post_composed trait ontology. Fixed in the Okra db, and made another commit to find or create the correct root term

chris263 (approved):

PR #5498, opened 2025-05-16
fixes for the image analysis tool

lukasmueller:
Description 
The image analysis tool configuration was moved to the conf files, but some data was still hard coded (trait data). This moves the trait data also to the config file.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


afpowell (changes requested):
I'm getting a "garbage after JSON object..." error. It looks like the braces right before "whitefly count" in the sgn.conf aren't quite right. I think one closing brace needs to be deleted.

afpowell (approved):
:+1:

PR #5514, opened 2025-05-27
Flexible headers in fieldbook file upload

bm743:
Description 
Allow headers in fieldbook file to be in any order. Files exported from fieldbook now work with the uploader

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


afpowell (changes requested):
It looks like the check of the ObservationUnitDbId values might be failing when it shouldn't. It passes verification if I change the values in that column to be the same as those for ObservationUnitName, but this is not what is exported from Field Book.

chris263 (approved):
It works!

afpowell (approved):
:+1:

PR #5585, opened 2025-07-03
Fixing report generation for large files

chris263:
The report generation process was intermittently failing to decode JSON output files. The error message indicated that the JSON was incomplete or malformed (e.g., unexpected end of string while parsing JSON string).

This happened because the code was attempting to read the JSON file as soon as it appeared on disk, before the report generation script had fully finished writing the file. Since the script was launched asynchronously via CXGN::Job, there was no guarantee that the file was complete at the moment it became visible.

Solution:
To ensure the JSON file is fully written before reading, the code was updated to execute the report script synchronously using system() instead of submitting it to CXGN::Job. This guarantees that the process has completed and the file is complete before any attempt to decode its contents.

This change eliminates race conditions and ensures consistent, valid JSON parsing.


chris263:
There is new PR in substitution

PR #5600, opened 2025-07-09
Update Report Engine

chris263:
It is in substitution on the PR #5585 

1- Supports running multiple report formats in a single request.
2 - Executes report generation asynchronously, allowing the user interface to remain responsive while reports are processed in the background.
3- Implements automatic error notification: if the report script fails, an email is sent to notify users of the error details.
4- Fixes phenotype QC report


ryan-preble (changes requested):
In Report.pm, you will need to manually update the status of the job record in the DB depending on what happens. Use:
$report_job_record->update_status("failed");
$report_job_record->update_status("finished");
where appropriate.

ryan-preble (approved):

Issue #5599, opened 2025-07-09
Search Wizard: Include Subplots


dwaring87:
Currently plots, plants, and tissue samples are supported (starting with the 2nd column) but subplots are not

Issue #5520, opened 2025-05-29
Wizard: loading selection from list  for tissue samples type is not working


isaak:
-- error message: 'List is not of an appropriate type.' This happens even after validating the list.  (From a meeting with Prasad)

PR #5647, opened 2025-07-30
adding subplot to wizard

chris263:
- Addin subplots to wizard
- Accept lists with tissue samples in wizard

Fixes: #5520
Fixes: #5599

### DIFF
--- a/db/00196/AddSubplotToMaterializedview.pm
+++ b/db/00196/AddSubplotToMaterializedview.pm
@@ -1,0 +1,1630 @@
+#!/usr/bin/env perl
+
+
+=head1 NAME
+
+AddSubplotToMaterializedview.pm
+
+
+=head1 SYNOPSIS
+
+mx-run AddSubplotToMaterializedview [options] -H hostname -D dbname -u username [-F]
+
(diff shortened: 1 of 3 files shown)


lukasmueller:
putting a list of type tissue_samples still seems to return that error, list type not supported

chris263:
1. I found an additional issue, which has now been fixed: the Trial Details page was creating the tissue samples list under the wrong type.
2. A typo (`tissue_sample` vs. `tissue_samples`) was causing the problem in the search wizard.
3. It is now possible to load a list with the `tissue_samples` type.

lukasmueller:
I created a subplot and a list of type subplots. When feeding the subplot list, I get again the error : "List is not of an appropriate type." For lists of type tissue_sample, this error no longer appears :-)

chris263:
@lukasmueller subplots lists are validated now!

chris263:
I've moved the db folder to 00198 to avoid any problems with new db patches.
Fixed subplot search.
Screenshot from 2025-09-15 10-11-46

alockrow:
I'm not getting that error anymore!

Issue #5689, opened 2025-09-03
identifier generation does not work under certain conditions


lukasmueller:
Identifier generation does not work if the username has a dash in it, as it runs grep and this creates an illegal regular expression.

lukasmueller:
The issue is the grep command on line 397:

```sub remove_element {
    my $self = shift;
    my $element = shift;

    my $h = $self->dbh()->prepare("DELETE FROM sgn_people.list_item where list_id=? and content=?");

    eval {
	$h->execute($self->list_id(), $element);
    };
    if ($@) {

	return "An error occurred while attempting to delete item $element";
    }

		eval {
    	my $q = "UPDATE sgn_people.list SET modified_date = now() WHERE list_id=?";
    	my $h1 = $self->dbh()->prepare($q);
    	$h1->execute($self->list_id());
		};

    my $elements = $self->elements();
    my @clean = grep(!/^$element$/, @$elements);
    $self->elements(\@clean);
    return 0;
}```

lukasmueller:
topic/fix_identifier_generation_issue

PR #5691, opened 2025-09-03
Fix issue with identifier generation

lukasmueller:
The username is stored in the json structure for lists of type identifier_generation. If the username contains a dash, the remove_element function crashes as it tries to use grep to remove elements, interpreting it as an illegal re range. Going the old-fashioned loop/match route instead that does not require regular expressions.

Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/List.pm
+++ b/lib/CXGN/List.pm
@@ -387,14 +387,26 @@ sub remove_element {
 	return "An error occurred while attempting to delete item $element";
     }
 
-		eval {
+    eval {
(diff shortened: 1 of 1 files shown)


lukasmueller:
fixes #5689

isaak (approved):

PR #5742, opened 2025-10-07
Compare lists

### LINKED ISSUE #5656: annoying things about lists
observed on CASSbase.

There are a few things about lists that annoy me. It's not super urgent, but maybe I'm not alone with my thoughts:
I sorted them by importance. 

- all items: if I want to add all items of a search result, it will only add the ones on the actual page! For example, if I search for an accession and get two pages of results, it will only add the open page to the list. It might be better to always have "all" search results shown and manually reduce the number to 10/50/100 if needed. 
- copy/paste (list manager): I would like to directly copy and paste things from a list. I always have to download, save, and open it, copy/paste, and then delete it again. 
- batch download/delete: I like the new feature with merging multiple lists, but it would be great if I could batch download or delete multiple lists at once.
- sorting of lists (list manager): They are now sorted by date (which is great), but also alphabetically. If I create multiple lists on the same day, they are sorted by name (within that day) and not by time. I would prefer to see them in the order that I created them. 
- multiple pop-ups: when I delete a list, I receive two pop-ups: First, the question if I really want to delete, and the second that it was deleted. Both times, I have to confirm. That is actually true for multiple situations on CASSbase, not just the list manager.

bm743:
Description 
Adds ability to see accessions that are shared/different between two selected lists

Fixes #5656 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/js/source/legacy/CXGN/List.js
+++ b/js/source/legacy/CXGN/List.js
@@ -264,6 +264,22 @@ CXGN.List.prototype = {
         return lists;
     },
 
+    compareLists: function(list_ids) {
+        return jQuery.ajax({
+            url: '/list/compare',
+            data: { 'list_ids': list_ids.join(',') },
+            method: 'GET',
+            success: function(response){
+                console.log("list1: ", response.only_in_list1);
+                console.log("list 2: ", response.only_in_list2);
+                if (response.error) {
+                    alert(response.error);
+                }
+            }
+        });
+    },
+    
+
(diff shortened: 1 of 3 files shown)


alockrow:
The button isn't showing up for me for some reason. Also, adding this function to the list groups test would be good!
image

PR #5766, opened 2025-10-27
Updates to transcriptomics upload

### LINKED ISSUE #5407: Updates to NIRS data uploader
Notes from Mckena:
- For the specific spreadsheet the following should be included based on our [data model](https://docs.google.com/spreadsheets/d/1WCTO0k79HyBQS5obOA6e4CasNNHedhDafNA5Z494VgQ/edit?gid=1024253684#gid=1024253684):
  **Required**
          - Sample ID: sample name is present already ✔️
          - Tissue Type:  Description of the plant part (e.g. leaf) or the plant product (e.g. resin) from which the sample was taken, in the form of an accession number from Plant Ontology
      - Wavelengths: Present already✔️
  **Optional**
          - Scan Timestamp: Timestamp specific to each individual scan, provided by some spectrometers
          - Comments: Present already ✔️
  
- Remove device_id from template: this should already be filled in the protocol section (or no? Discuss w/group)

- Ensure everything listed as optional is actually optional. Or specify that the df being uploaded needs to include all optional column headers as well.

- Site freezes after "Second cell must be 'device_id'" error in spreadsheet verification

- I don’t think this “Data Level” drop down is necessary on final page of uploader, there are no options other than Samples, it can be removed

bm743:
Description 
Add protocol fields and data matrix headers to match BrAPI spec, allow verifying uploaded files multiple times if there is an error

Fixes #5407 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/mason/tools/high_dimensional_phenotypes/upload_high_dim_spreadsheet_dialogs.mas
+++ b/mason/tools/high_dimensional_phenotypes/upload_high_dim_spreadsheet_dialogs.mas
@@ -13,6 +13,7 @@ my @layout_options = split ",", $c->config->{layout_options};
 my @mapping_software_options = split ",", $c->config->{mapping_software_options};
 my @sequencing_platform_options = split ",", $c->config->{sequencing_platform_options};
 my @na_extraction_method_options = split ",", $c->config->{na_extraction_method_options};
+my @sequencing_center_options = split ",", $c->config->{genotyping_facilities};
 </%perl>
 
 <div class="modal fade" id="upload_high_dim_phenotypes_spreadsheet_dialog" name="upload_high_dim_phenotypes_spreadsheet_dialog" tabindex="-1" role="dialog" aria-labelledby="addHighDimDataDialog">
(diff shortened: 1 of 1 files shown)


titima15:
TranscriptomicsSpreadsheetCSV is the old format plugin. Should we update to the generic plugin? How about test update?

mason/tools/high_dimensional_phenotypes/upload_high_dim_spreadsheet_dialogs.mas
```diff
@@ -167,6 +167,76 @@ $high_dimensional_phenotype_type => undef #either NIRS, Transcriptomics, Metabol
...
+
+                            <div class="form-group">
+                                <label class="col-sm-3 control-label">Read Length: </label>
+                                <div class="col-sm-9" >
+                                    <input class="form-control" id="high_dim_data_upload_transcriptomics_read_length_select" type = "text" placeholder="e.g. 5x depth with 150 bp reads">
+                                </div>
+                            </div>
+
+                            <div class="form-group">
+                                <label class="col-sm-3 control-label">Nucleic Acid Extraction Method: </label>
```
afpowell: It looks like the extraction method field may be duplicated here?

afpowell (approved):
:+1:

PR #5778, opened 2025-10-31
Fix loop condition for stock types iteration

bvsam:
Description 

## Bug
- Line 35 contains an off-by-one error that causes an infinite loop when:
  - Stock table is empty, OR
  - No "accession" stock type exists
- Using `<=` instead of `<` in the loop condition causes accessing  `$stock_types->[$i]` where `$i` equals the array size (out of bounds). This triggers Perl autovivification, which creates new array elements, increasing the array size, which makes the loop condition true again, creating an infinite loop that consumes unbounded memory.

## Change
- Change `$i<= scalar(@$stock_types)` to `$i< scalar(@$stock_types)`
- Prevents memory leaks caused by Perl's autovivification in cases where the stock table is empty or no "accession" stock type exists.

## Testing
- I've tested these changes locally (within a Docker container based off [breedbase_web](https://hub.docker.com/r/triticeaetoolbox/breedbase_web)) and I can confirm they fix the bug.
  - with empty stock table: page loads successfully
  - with populated stock table: existing functionality preserved

I encountered this issue when using [TriticeaeToolbox's sgn](https://github.com/TriticeaeToolbox/sgn), but this repository seems to be the upstream for their fork and the issue exists in both repositories.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/mason/search/stocks.mas
+++ b/mason/search/stocks.mas
@@ -32,7 +32,7 @@ use CXGN::Page::FormattingHelpers qw / conditional_like_input_html simple_select
 
 my $accession_cvterm_id;
 
-for (my $i=0; $i<= scalar(@$stock_types); $i++) {
+for (my $i=0; $i< scalar(@$stock_types); $i++) {
      if ( $stock_types->[$i][1] eq "accession" ) {
      	$accession_cvterm_id = $stock_types->[$i][0];
 	last();


lukasmueller:
Thank you so much for submitting this pull request! This is an issue that has bothered me for a while. Good to see the fix :-) !

bm743 (approved):

PR #5783, opened 2025-11-05
Accession Transform: return ids in same order as original name list

### LINKED ISSUE #5782: Accession List Items assigned wrong IDs in Search Wizard
In the Search Wizard, when you select a list of Accessions from the first column, you can click on the list item to open the stock detail page for that accession.  The links are to the wrong accession.

For example, on cassavabase, select `yellow_GWAS23_accessions` from the first column.  If you click the item named `IITA-TMS-IBA30572` it opens the stock detail page for `IITA-TMS-IBA000070`.

This only seems to be a problem for lists of accessions.  Lists of trials, etc seem to link to the correct item.

It looks like the stock ids are not being assigned to the matching stock name.  The results of the accessions 2 accession_ids transformation look to be sorted by stock id, which is not the original order of the list.

dwaring87:
Description 

This updates the accessions2accession_ids list transformation to return the found ids in the same order as the original list names.

Fixes #5782

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/List/Transform/Plugin/Accessions2AccessionIds.pm
+++ b/lib/CXGN/List/Transform/Plugin/Accessions2AccessionIds.pm
@@ -43,9 +43,7 @@ sub transform {
 
          # add map of found name --> found id
          for my $i (0 .. $#found_names) {
-            my $n = $found_names[$i];
-            my $i = $found_ids[$i];
-            $found_hash{$n} = $i;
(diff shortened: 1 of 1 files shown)


afpowell (approved):
:+1:

PR #5818, opened 2025-12-08
Speed up plots/subplots/plants experimental design tables

### LINKED ISSUE #5811: Plot list on trial detail slow
It takes a long time for certain sites to load the plot list on the trial detail page.
(as reported by Guillaume)

ryan-preble:
Description 

Changes the query that makes the plots/subplots/plants tables to include all data so no time is wasted collecting data through object methods. Also tries to optimize the queries that were slow by aggregating subqueries and selecting by stock ID before any joins.

I tried testing on cassava devel. Trials with 6000+ accessions take about 20 seconds to load, but the page itself also took about that long to load. Trials with less than 1000 stocks took a few seconds. 2000 plants took 11 seconds to load. Not sure how fast it will run on a production site. 

Fixes #5811 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


dwaring87 (approved):
The speed is greatly improved! I tested on some trials with 700-800 plots and it went from loading ~17 plots/sec to ~670 plots/sec 🚀

PR #5843, opened 2026-01-08
Spatial layout download improvements

### LINKED ISSUE #5795: Spatial layout dowload improvement
The ability to download the spatial layout of a trial is great, but the current export only displays the line, which isn’t sufficient when working with replicates.

Suggestion:
It would be very helpful to have the option to choose which fields are shown in the spatial layout. For example observationunit_name, plot_ID, line, tags (see issue: #5794), etc. 

This would make the exported layout far more informative and usable, especially for trials with multiple replicates.

bm743:
Description 
Adds options to choose to include accession name, observation unit name, and plot ID for each plot when downloading spatial layout

Fixes #5795

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


dwaring87 (changes requested):
I also noticed that when accession name is checked, null is printed in the output file when the stock type in the plot is a cross.

dwaring87 (approved):

mason/breeders_toolbox/trial/phenotype_heatmap.mas
```diff
@@ -483,6 +483,46 @@ $has_plant_entries => undef
...
+            </div>
+            <div class="modal-body">
+                <div class="container-fluid">
+                    <form class="form-horizontal" role="form" method="post" >
+                        <div class="well">
+                            <div class="form-group">
+                                <div class="col-sm-3 control-label"><strong>Information to include: </strong></div>
+                                <div class="col-sm-9" >
+                                    <div class="checkbox">
+                                        <label><input type="checkbox" id="accession_name" value="accession_name" checked> Accession Name </label>
```
titima15: For the user interface, if trial stock type is cross, could you please use 'Cross/Accession Name' label instead of 'Accession Name'?

PR #6013, opened 2026-04-02
Fix trial title to top of page

### LINKED ISSUE #5980: Feature Request: Keep trial name fixed on Trial Detail Page
On the Trial Detail Page, the trial name scrolls out of view, making it difficult to identify which trial is currently being viewed.

This also applies to accession and vector detail pages, though it is particularly noticeable for trials, as we often work with multiple trials in parallel across different windows.

bm743:
Description 

Fixes #5980 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/mason/breeders_toolbox/trial.mas
+++ b/mason/breeders_toolbox/trial.mas
@@ -72,7 +72,9 @@ $project_id => undef
 
 <br/>
 
-<& /page/page_title.mas, title=>"Trial detail for ".$trial_name &>
+<div style="position: sticky; top: 55px; background: white; z-index: 1000;">
+  <& /page/page_title.mas, title=>"Trial detail for ".$trial_name &>
+</div>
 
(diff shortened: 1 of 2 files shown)


ryan-preble (changes requested):
I would add a little bit of border and padding around the trial name to make it prettier. The way it makes the entire white bar float over the page without any border/visual break is a little jarring. Rounded corners to match other page components would also be clean

Also, the way that the rest of the page is still visible in a narrow strip above the title bar could get cleaned up

lukasmueller:
yes, agree with that!

lukasmueller (approved):
:-)

ryan-preble (approved):

PR #6015, opened 2026-04-02
Intercropping: Support Multiple Trait Ontologies

### LINKED ISSUE #5309: Intercropping: Support multiple trait ontologies
For our oat/pea intercropping trials, the trait observations include some oat-specific traits and pea-specific traits.  For now, we've been adding the pea-specific traits to the composed trait ontology (to keep the oat trait ontology clean and in-sync with cropontology.org).  We'd like to be able to have additional trait ontologies loaded in the database (oat trait ontology, pea trait ontology, composed trait ontology) so each crop has a separate trait ontology.

Requirements:

- Be able to load additional trait ontologies into the database (should already be possible?)
- Tag the additional trait ontologies as a `trait_ontology` in the `cvprop` table
- Ensure the additional trait onologies are displayed on the Search > Traits page
- Ensure that traits from the additional trait ontologies can be used when uploading trait observations
- Ensure that traits from the additional trait ontologies are included in the BrAPI endpoints (and field book)

dwaring87:
Description 

This adds some additional intercropping support:

- DB Patch: update the traits view to support multiple trait ontologies (when there are multiple `VARIABLE_OF` terms in the database) - [DB Patch Diff](https://www.diffchecker.com/e2Dw82GX/)
- TrialLayout: add intercropped accessions to the layout design (`intercrop_accessions` key for each plot) and list of accessions (in `accession_names` key of layout)
- Onto: make DB search by name case-insensitive
- Tests: add specific tests to check for intercropping support (in trial upload, trial design, and phenotype matrix)
- Docs: add intercropping subsection to the managing field trials chapter

Fixes #5309

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


ryan-preble (changes requested):
There are a number of places where the `trait_ontology_db_name` config key is used and treated like it can only be a single value. I think it would be good to check that it is being correctly parsed as though it could have multiple values. 

Also, this begs the question: is the only way to add a second trait ontology by manually inserting a row into cvprop? Perhaps there should be a way to make updates between sgn_local.conf and cvprop coherent?

Alternatively, maybe the places that use `trait_ontology_db_name` need to be shifted away from using the config and instead rely on cvprop for the actual identities of the trait ontolgies

lukasmueller:
Should be querying the database from the cvprop table. 
Ontologies should be user loadable (ticket?)

ryan-preble (approved):

ryan-preble:
Something that I realized while looking at #6044 : the trait designer assumes only one trait ontology for autocompleting names and showing the embedded ontology browser. Same for downloading the trait ontology - would need to be changed to download multiple ontologies, either in series or all at once. This is potentially a bit of a breaking change

lukasmueller:
Yes, it should consider all the trait ontologies listed in cvprop.

lukasmueller (approved):
:-)

PR #6026, opened 2026-04-09
fix wizard stuck updating

### LINKED ISSUE #6025: problem with materialized view refresh
In the wizard, the "Update Wizard" function will randomly not work because the database is stuck in "currently_updating" status. this may be caused when bin/refresh_matview.pl and bin/ refresh_materialized_markerview.pl fail when uploading data.

refresh_matview.pl
1.  run as a transaction (autocomit->0) can fail with "REFRESH MATERIALIZED VIEW" command. If transaction fails the database will remain in currently_updating = TRUE
2. when run with concurrent option it is also dangerous to run in transaction

refresh_materialized_marverview.pl
1. run as a transaction (autocomit->0)  can fail with "REFRESH MATERIALIZED VIEW" command. .
2. usr try catch instead of eval to get better error checking

ClayBirkett:
wizard stuck in updating state may be caused by loading data while there is a materialized view update in process

Improves error handling during materialized view refresh
uses pg_advisory_lock to prevent execution if there is a currently running update
improves logic when updating "currently_refreshing" in database

closes #6025 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/bin/refresh_materialized_markerview.pl
+++ b/bin/refresh_materialized_markerview.pl
@@ -28,15 +28,16 @@ use strict;
 use warnings;
 use Getopt::Std;
 use DBI;
+use Try::Tiny;
 
 our ($opt_H, $opt_D, $opt_U, $opt_P);
 getopts('H:D:U:P:');
 
 print STDERR "Connecting to database...\n";
 my $dsn = 'dbi:Pg:database='.$opt_D.";host=".$opt_H.";port=5432";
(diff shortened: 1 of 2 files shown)


ryan-preble (approved):
I was unable to break it!

lukasmueller (approved):
:-)

PR #6034, opened 2026-04-10
Add markers prot

ClayBirkett:
add option to append markers to genotype protocol during VCF upload

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [x] New feature
  - [ ] User-Facing Change

### DIFF
--- a/bin/load_genotypes_vcf_cxgn_postgres.pl
+++ b/bin/load_genotypes_vcf_cxgn_postgres.pl
@@ -130,6 +130,10 @@ my $add_accessions = 0;
 if ($opt_a){
     $add_accessions = 1;
 }
+my $add_markers = 0;
+if ($opt_m){
+    $add_markers = 1;
+}
 my $include_igd_numbers = 0;
 if ($opt_z){
     $include_igd_numbers = 1;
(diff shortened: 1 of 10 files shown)


ClayBirkett:
The feature of adding new markers to a genotype protocol works only for VCF files. It does not work for KASP, SSR, TASSEL, or Intertek.

ClayBirkett:
I have added code so that you can append new markers for KASP and Intertek file types. I have tested this with Intertek and it seems to work. The same code is used for KASP markers

titima15:
image
Shouldn't we see a list of additional markers on this page?

ClayBirkett:
The missing markers are listed after you do the upload. I can change the wording to "Add missing markers?:" If it is run with "No" selected then you get error with list of missing markers. If it is run with "Yes" selected then you it automatically adds them during upload. This is similar the to "Add missing accessions?" option that we used to have.

titima15:
Yes, I think it's better to change the wording. When I saw "These missing markers", I expected to see a list of missing markers. I went through the whole process with vcf upload, it works as expected.

titima15:
I think it will be even better if you can see the list of additional markers before proceeding to the submit step.

ClayBirkett:
The way it works is you submit the upload and it tells you which markers are not in the protocol. Then you go back and check the box to "add missing markers". It is basically the same with accessions. You submit the file and it tells you which accessions are not loaded, you load these, then try again.

titima15 (approved):

lukasmueller (approved):
:-)

PR #6038, opened 2026-04-13
Marker Metadata Improvements

dwaring87:
Description 

This updates the marker metadata file upload and display table on the genotyping protocol page

- When uploading a marker metadata file:
     - if the marker name does not exist in the `phenome.locus` table: the locus is created and the marker metadata added
     - if the marker name exists and is associated with the selected protocol in the `phenome.locus_geno_marker` junction table: a warning message is returned and the the locus and metadata are updated if warnings are ignored
     - if the marker name exists and is NOT associated with the selected protocol in the `phenome.locus_geno_marker` junction table: an error message is returned and the existing name cannot be used
 - Updates to the marker metadata data table on the protocol detail page:
     - added a references column to display and link to associated ontology terms and/or external dbs
     - added an edit column to allow a curator to edit existing metadata or remove a single locus and saved metadata
     - added a delete all marker metadata option (curator only)
 - Add mech tests to test adding, fetching, updating, and deleting marker metadata 

image

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Genotype/ParseUpload.pm
+++ b/lib/CXGN/Genotype/ParseUpload.pm
@@ -68,6 +68,14 @@ has 'parse_errors' => (
     predicate => 'has_parse_errors',
 );
 
+has 'parse_warnings' => (
+    is => 'ro',
+    isa => 'HashRef',
+    writer => '_set_parse_warnings',
+    reader => 'get_parse_warnings',
+    predicate => 'has_parse_warnings',
+);
+
 has '_parsed_data' => (
     is => 'ro',
     isa => 'HashRef',
--- a/lib/CXGN/Genotype/ParseUpload/Plugin/MarkerMetadata.pm
+++ b/lib/CXGN/Genotype/ParseUpload/Plugin/MarkerMetadata.pm
@@ -40,7 +40,7 @@ sub _validate_with_plugin {
     my $parsed_data = $parsed->{data};
 
     my @markers = @{$parsed->{values}->{'Marker'}};
-    my @aliases = @{$parsed->{values}->{'Alias'}};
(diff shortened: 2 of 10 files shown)


lukasmueller (commented):
>if the marker name exists and is NOT associated with the selected protocol in the  phenome.locus_geno_marker junction table: an error message is returned and the >existing name cannot be used

Isn't is theoretically possible that two markers have the same name between two protocols and are associated with different loci?

dwaring87:
> > if the marker name exists and is NOT associated with the selected protocol in the  phenome.locus_geno_marker junction table: an error message is returned and the >existing name cannot be used
> 
> Isn't is theoretically possible that two markers have the same name between two protocols and are associated with different loci?

Each marker in the protocol is linked to a locus in the `phenome.locus` table, which needs to have a unique name.  If you have two markers with the same name that are associated with different loci then you can use the alias column when uploading the marker metadata to give them unique names.  The alias is used as the name for the `phenome.locus` entry.

lukasmueller:
Great, that makes sense.

titima15:
@dwaring87 There is a bug in the validation step. If there is no Alias column, it gives an error (Can't use an undefined value as an ARRAY reference: line 43)

dwaring87:
> @dwaring87 There is a bug in the validation step. If there is no Alias column, it gives an error (Can't use an undefined value as an ARRAY reference: line 43)

@titima15 Thanks for catching that.  It should be fixed now.

titima15 (approved):
👍

lukasmueller (approved):
:-)

Issue #5808, opened 2025-11-26
Enable Curator to Change Phenotypes & Tabular View


MercedesThieme:
I’m looking for a way to edit specific phenotypic values after they’ve been uploaded to CASSbase. This also relates to an older request (#5464), where I mentioned the need for an Excel-like, tabular view of phenotype data.

In many cases, I need to see traits in relation to one another (e.g., does a small root value correlate with a small shoot value or could it be a typo/outlier?). Currently, CASSbase only displays phenotypic values one trait at a time in a difficult-to-read format. The alternative is to download the data just to view it in a table, but that defeats the purpose—people will rely on their downloaded spreadsheets instead of using CASSbase directly.

A built-in, editable (by curator!), spreadsheet-style interface for phenotype data would significantly improve usability and data consistency.

Issue #5722, opened 2025-09-18
tiny broken things


MercedesThieme:
I found multiple small things that are not working. Not important to fix, just for your information.

**Trial Detail Page**
1. Phenotype Summary Statistics / Download Trial Data only works if CSV is selected. It throws an error if XLSX is chosen.
2. The symbol for "Compute new phenotype" looks like a save icon, which is confusing.
3. Why is there no save option on the trial detail page? To download phenotype data, one has to go through Phenotype Summary Statistics. That’s not very intuitive.
4. Phenotype Summary Statistics / Export to Expression Atlas: I don’t think this is the right location for this option. Also, I’m not sure how useful it is to send phenotypes to EA. I’d assume that would be a separate project. Even if I had transcriptome data for a trial, they wouldn’t be shown together, or would they?
5. Phenotype Summary Statistics / Histogram: Nice little feature, but without adjustable bin width, it’s almost useless (depending on the trait).
6. Analysis: No option to select plants (instead of plots) for correlations. ANOVA is not supported for greenhouse designs.

**Label Designer**
Big fan of this tool, but I think it could be even better.
1. It’s not obvious which parameters were used in a saved label design. This is especially an issue with the QR code. To be sure, I have to recreate and arrange the elements each time. Maybe saving placeholders in the template would help, and each element has to be defined per trial.
2. Added text is center-aligned, which can cause issues if a long name overlaps with the QR code. Left alignment would prevent this (assuming the QR code stays on the far left).

**Field Trial Design**
Like I’ve mentioned before, I really need the option to add plants after a trial has been created. (This one is actually important to fix, in my opinion.)

lukasmueller:
I think most of these issues were addressed. Can this issue be closed?

PR #6046, opened 2026-04-14
Topic/phenotype UI edits

### LINKED ISSUE #5989: Enable easy editing of phenotype data in CASSbase
observed on CASSbase & Fieldbook. 

Typos during phenotyping are common, but correcting them is difficult. Errors are often only noticed after upload to CASSbase, and fixing them later in Fieldbook changes timestamps. CASSbase currently doesn’t allow edits, leaving trial deletion/recreation as the only workaround (curator-only and inefficient).

**Suggestion:**
Add an Excel-like table view in CASSbase to let trial owners and curators easily review and edit data, ideally with change tracking.

Related issues: #5464, #5808

ryan-preble:
Description 
Adds the ability to edit individual phenotype values from the trial detail page. People with permission to edit a trial can go to phenotype raw data and edit a single row at a time. Also changes the glyph for compute new phenotypes section (not the best because glyphicons is limited, would have to install another icon library to get more)

Closes #5989 #5808 
Addresses a point in #5722 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [x] New feature
  - [ ] User-Facing Change


lukasmueller (commented):
* The storage function errors out when there is no date provided
* When no operator exist in the original record, it errors out because undef does not seem to be allowed for operator.

dwaring87 (approved):

lukasmueller (approved):
:-)

PR #6059, opened 2026-04-22
QuickSearch improve error checking

ClayBirkett:
website was crashing from bot use of quicksearch
fixed a few problems

Checklist 
- [x] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/SGN/Controller/QuickSearch.pm
+++ b/lib/SGN/Controller/QuickSearch.pm
@@ -191,20 +191,20 @@ sub execute_predefined_searches: Private {
     my $sp_person_id = $c->user() ? $c->user->get_object()->get_sp_person_id() : undef;
     # execute all the searches and stash the results
     for my $search_name ( sort keys %searches ) {
-	print STDERR "performing quick search for $search_name (". Dumper($searches{$search_name}).")...\n";
-         my $search = $searches{$search_name};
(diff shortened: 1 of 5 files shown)


ryan-preble (approved):
I have often used both eval and Try::Tiny to do error handling, but I hear that Try::Tiny is better. Should we commit to using one or the other for new code or does it not matter? Does Perl wizard Lukas have any input?

lukasmueller:
> I have often used both eval and Try::Tiny to do error handling, but I hear that Try::Tiny is better. Should we commit to using one or the other for new code or does it not matter? Does Perl wizard Lukas have any input?

Try::Tiny is just syntactic sugar on an eval, as far as I know. So they two should be equivalent, but Try::Tiny is a bit easier on the eyes, so potentially we should prefer it for that reason.

lukasmueller:
Maybe the quick search should be behind a login? User configurable in sgn_local.conf?

lukasmueller (approved):
Some of the searches are exact match only. Should there be a substring match option, or should that be the default?

:-)

lukasmueller:
Approved this but it really awaits being put behind a login with a config setting

PR #6087, opened 2026-05-01
Fix exif verification

bm743:
Fixes exif error occurring when observation_variable_name field does not have both the trait name and ID  in this format trait_name|trait_ID. Just the trait name will not work as well

Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


chris263 (approved):

chris263 (approved):

lukasmueller (dismissed):
It would be better to use the cvterm_id of the trait to retrieve that cvterm object. The name might not be unique across several ontologies.

dwaring87 (approved):

PR #6093, opened 2026-05-05
Environment stratification tool

chris263:
Description 

This PR adds the new Environment Stratification tool to Breedbase.

  The tool allows users to select a dataset and trait, run environment stratification analysis, and review compatible environment groups based on genotype-by-
  environment interaction. Results are returned through a new AJAX controller, processed by an R backend script, and displayed in a dedicated Mason/JavaScript
  interface.

  Main Features

  - New Environment Stratification page and UI.
  - Dataset and trait selection using existing Breedbase dataset tools.
  - R-based environment stratification analysis.
  - Pairwise environment compatibility testing.
  - Group summary output for statistically compatible environments.
  - ANOVA output tab with automatic experimental design detection.
  - Ungrouped environment output.
  - Interactive map showing trial locations and environment groups.
  - Hover tooltips on map markers showing location and trial name.
  - XLSX downloads for result tables.

  ANOVA / Design Handling

  The R script detects the available experimental design information and builds the ANOVA model accordingly.

  Supported fixed-effect ANOVA structures include:
  - CRD
  - RCBD
  - Incomplete block / alpha-lattice
  - Augmented / block-only
  - Row-column / augmented row-column

Row and column terms are only used when the trial design is explicitly row-column/spatial. They are ignored for CRD and RCBD even if coordinate columns are present.

Screenshot from 2026-05-05 14-36-05

image

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ x] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Dataset.pm
+++ b/lib/CXGN/Dataset.pm
@@ -1300,9 +1300,6 @@ sub calculate_tool_compatibility {
         'Correlation' => {
             'compatible' => 0
         },
-        'Dataset Analysis' => {
-            'compatible' => 0
-        },
         'Environment Stratification' => {
             'compatible' => 0
         },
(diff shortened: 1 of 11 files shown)


ryan-preble (changes requested):
The tool works well and shows up properly on the job submissions table - good work!
It would also be good to add some lines to the tool compatibility function to display dataset compatibility with the analysis.

lib/SGN/Controller/AJAX/EnvironmentStratification.pm
```diff
@@ -0,0 +1,324 @@
...
+
+BEGIN { extends 'Catalyst::Controller::REST' }
+
+__PACKAGE__->config(
+    default   => 'application/json',
+    stash_key => 'rest',
+    map       => { 'application/json' => 'JSON' },
+);
+
+sub shared_phenotypes : Path('/ajax/environment_stratification/shared_phenotypes') : {
```
lukasmueller, line 24: Should these functions be only available to logged in users?

lukasmueller (commented):
check comment about login requirement

:-)

chris263:
I have added dataset compatibility check plus login is required to use this tool.

lib/CXGN/Dataset.pm
```diff
@@ -1344,6 +1350,14 @@ sub calculate_tool_compatibility {
         }
     }
 
+    foreach my $analysis_name ("Dataset Analysis", "Environment Stratification") {
```
ryan-preble: This is different from how all the other tools determine compatibility. It seems like this only requires a dataset to have trials and traits, but does not impose any requirement on the number of observed phenotypes. The other tools have a "warn" condition that explains the dataset meets the minimum requirements but may have low N, making results unreliable. The "dataset requirements" key is not present in any other tool. It is also unclear to me what the "Dataset Analysis" tool is and why it has the same requirements as environment stratification.

chris263: I think this part should be flexible for breeders to decide how they are mounting the dataset to construct zones. They just need trials and at least a trait. Make it too restrict I think can cause more complains.

ryan-preble (approved):

lukasmueller (approved):
:-)

Issue #6096, opened 2026-05-07
Create a post composed traits plotting tool


lukasmueller:
Enter a number of composed terms that have been post composed with the time ontology, and generate a bar graph.

As suggested in the discussion of #6091.

ktmeaton:
I just discovered this is kind of already possible? In the **Analysis** section of a field trial, I select **Get Time-Series Plots for Accessions on Saved Measurements**. And then I pick my traits that have a time component and accessions, and then voila, time series plot!

<img width="1059" height="627" alt="Image" src="https://github.com/user-attachments/assets/a70761ee-4800-4186-8c00-0e6152e91cc1" />

<img width="1546" height="834" alt="Image" src="https://github.com/user-attachments/assets/6a52a2d9-a7a1-4047-9d30-4567963acedc" />

The formatting of the final result is not ideal, as it extremely zoomed in. I'm really surprised this works even with years as the time component (#6100). It might be slightly nicer to have this as an interactive javascript component, but I'm very excited the basic functionality already exists!

ryan-preble:
Neato mosquito - I did not know that existed. Instead of changing the repetitive measurements section, I will focus on updating this tool to return the raw data and reconstruct a semi-interactive D3 plot rather than returning the R ggplot output

PR #6100, opened 2026-05-11
Add year to cxgn time ontology

### LINKED ISSUE #6097: Adding year to the cxgn time ontology
We have some old historical field trials that are coming up on 50 years of measurements soon. If I contributed a pull request to add `time in years` to the time ontology, would that be a welcome contribution? If it's not useful to your other breeding programs, I can always keep it as a custom patch for our database. This is how I'm using it for composing traits and treatments:

Image

ktmeaton:
Description 

This pull requests adds time in  years (`year 1` to `year 100`) to the cxgn time ontology. In the config file, it can be enabled for compose traits and treatments by using the `tiy` config (time in years). I couldn't see any existing selenium tests for the compose interface, so I added some minimal tests under the `tools` category.

- Resolves #6097

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [x] New feature
  - [ ] User-Facing Change


ktmeaton:
Failing test 44 of `_BrAPIv2_phenotyping.t`, related to brapi v2 traits. Possibly the new time components are throwing off a check of traits in the db.

ktmeaton:
Yes, that's exactly what the problem was. I've now added 101 new 'traits' to the db, so the brapi expected count needs to be updated.

ktmeaton:
CI passing, ready for review!

db/00205/AddTimeOntologyYear.pm
```diff
@@ -0,0 +1,95 @@
...
+    print STDOUT "\nExecuting the SQL commands.\n";
+    my $schema = Bio::Chado::Schema->connect( sub { $self->dbh->clone } );
+
+    print STDERR "INSERTING CV TERMS...\n";
+    my $is_a_cvterm_id = SGN::Model::Cvterm->get_cvterm_row($schema, 'is_a', 'relationship')->cvterm_id();
+    my $time_ontology_rs = $schema->resultset("Cv::Cvterm")->find({name => "Time"});
+
+    # Create the root cvterm
+    # next available dbxref accession in time ontology is '0000481'
+    my $dbxref_accession = 481;
```
ryan-preble: I'm a little wary of hardcoding the next highest accession number in an ontology - I will need to fish around to see if other databases have messed with their time ontologies in such a way that it would break this

ktmeaton: That's a very good point, thank you. Initially I'd thought of using the `autocreated: year 1` accession for the `dbxref` accession. But then I thought this might be inconsistent in a bad way, since all the other time ontology accessions have hard-coded numeric accessions. Another option might be to put it in its own ontology (ex. `YEAR:000001`). But that might also be consfusing if year is the only time unit to have its own ontology.

I'm curious to see what you find out in the other databases!

ktmeaton: Hmm, maybe putting year in it's own time ontology was a bad suggestion. I imagine that will over complicate things downstream to have time in multiple ontologies (ex. plotting post-composed time traits #6096).

ryan-preble: I think your general structure of placing a new root term in the TIME ontology and then assigning that root term to be a TIY composable ontology is the right approach. But I checked and it seems other breedbases have added additional time terms to their DB that throws off the accession numbers. I think the best approach would be to solve for the next highest accession number in the db patch and use that to form the root term, then start adding new accession numbers from that starting point. Changing sgn.conf to use 481 makes sense because sgn_local.conf can override that value for all the databases that have modified their TIME ontologies away from the default. We will just have to be careful to set the correct values in sgn_local.conf wherever relevant

Also, one more thing: I am pretty sure cvterm create_with() will ignore the definition parameter. When I wrote a dbpatch that created new ontologies, I had to manually update the definition after making it with create_with().

lukasmueller:
Year separate would be complicated. Hardcoding might be fine

Get Outlook for iOS<https://aka.ms/o0ukef>

lukasmueller:
With hardcoding, I mean you predetermine what the IDs are, you don't let the script do it. That might lead to complications if the IDs are then not the same between systems.

lukasmueller:
Start at 0005000 for the year terms or something like that so that they are likely available on all sytems

db/00205/AddTimeOntologyYear.pm
ktmeaton: I've now changed the definition from inside create_with() to manual afterwards. Jumping over to Lukas' responses to discuss auto-detection of accession vs hard-coding a much larger root number.

ktmeaton:
Based on Lukas' comments, I've started the new time in years root to be at 0005000, with new entries going up to (and including 0005100) . @ryan-preble do you think this namespace is going to be compatible with most of the databases you've looked at so far?

sgn.conf
ryan-preble: Make sure tiy shows on this line too

ryan-preble, line 28: Question for @lukasmueller : do we want Time In Years to be available by default or suppressed until this line is overridden in sgn_local.conf?

ktmeaton: Ah I missed that, thanks! Fixed in latest commit. I'm glad the CI tests correctly failed when I had that value set incorrectly.

sgn.conf
```diff
@@ -31,6 +31,7 @@ composable_cvterm_format concise
 composable_variables 1 #display only variable terms in the post composing tool
 composable_tod_root_cvterm "time of day|TIME:0000001"
 composable_toy_root_cvterm "time of year|TIME:0000005"
+composable_tiy_root_cvterm "time in years|TIME:0000481"
```
ryan-preble: This needs to get bumped to 5000

ktmeaton: Thanks, just updated now.

ktmeaton:
Ah yes, CI is failing because the compose test is searching for specific time accession numbers to navigate the cvterm ontology tree in the UI. I will fix that test next.

ryan-preble (approved):

lukasmueller (approved):

PR #6103, opened 2026-05-12
Explicit defined checks for new unit values to permit '0'

### LINKED ISSUE #6102: New unit scale will not persist '0' values
1. Create a new observation variable by clicking "Add A New Observation Variable To the Database" from the Search->Traits page
2. At the "Scale" step, add a new scale
3. Set any of the numeric properties, such as minimum, maximum, or default value, to "0"

We should expect these "0" values to be persisted in the `cvtermprop` table, but they are not.

grapereader:
Description 

Fixes https://github.com/solgenomics/sgn/issues/6102

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Onto.pm
+++ b/lib/CXGN/Onto.pm
@@ -365,7 +365,7 @@ sub store_observation_variable_trait_method_scale {
     my $new_scale_name = shift;
     my $new_scale_definition = shift;
     my $new_scale_format = shift;
-    my $new_scale_minumum = shift;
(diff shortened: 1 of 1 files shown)


lukasmueller (approved):
:-)

bm743 (approved):

PR #6105, opened 2026-05-12
Explicit defined checks for exporting phenotype values to permit '0' 

### LINKED ISSUE #6104: Phenotype exports change values with '0' to an empty string
When exporting traits from the database, phenotypes with a value of 0 are being exported as an empty string. I've seen this consistently across export functions, including the Wizard. I think this is similar to #6102, in which checking if a value is present does not test for the edge case when it is a falsey value (ex. 0).

ktmeaton:
Description 

This PR fixes a bug in which phenotypes with a recorded value of '0' are exported from the database as an empty string.

- Resolves #6104 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
:-)

bm743 (approved):

PR #6128, opened 2026-05-29
Allow plants/subplots to be deleted or added again

### LINKED ISSUE #6033: Add / modify subplots or plants
Could we add the possibility, at least for curators, to add / delete / modify subplots or plants, even if some have already been entered for the trial ?

ryan-preble:
Description 
Allows curators to delete plants or subplots and to add additional plants or subplots to plots/subplots that already have entities. 

Closes #6033 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller (approved):
both deletion and addition worked. I didn't test all the different addition methods.

lukasmueller (commented):
Using the first option, "Adding plants", errored out when trying to add 4 plants per plot that already had 2 plants per plot. Not sure if that's by design.
Adding one plant by name with an excel sheet worked.

ryan-preble:
> Using the first option, "Adding plants", errored out when trying to add 4 plants per plot that already had 2 plants per plot. Not sure if that's by design. Adding one plant by name with an excel sheet worked.

I checked and could only reproduce this issue when adding plants in such a way that the coordinate system breaks down. Adding plants with unspecified coordinates to a plot with existing coordinates works - they just get no assigned coords. But it will break if the rows/columns specified can't accommodate extra plants AND the existing plants. I'll add a bit of explanatory text

titima15 (approved):

PR #6171, opened 2026-06-22
modify updated markers during upload

### LINKED ISSUE #6167: modify markers in genotype protocol
We sometimes get genotype data for an existing protocol where the markers are defined differently than the existing protocol.  Examples of this are
1. additional alternatives alleles
2. null alleles which can be represented as a deletion, so GT is 0/2 and ref/alt is A/T,-
The null allele is considered informative so I don't want to replace it with missing ./.

I have code to modify the existing markers, is this useful to BreedBase. The code allows update of ID, ref, alt but I expect it to only be used to update the alt column.

ClayBirkett:
If markers are changed (for example the alt field) then provide option to update existing protocol

fixes #6167 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [x] New feature
  - [ ] User-Facing Change


ClayBirkett:
The previous code lib/SGN/Controller/AJAX/GenotypesVCFUpload) iterates over every key in the uploaded marker's $new_marker_details hash and compares each one against the same key in the stored marker. So the fields compared are whatever keys happen to be present in the new marker data hash for each marker. Typically, for a CXGN::Genotype::Protocol marker, these would include things like chrom, pos, alt, ref, qual, filter, and info — but the code itself is field-agnostic and will compare whatever is there.

In this code I have limited it to chrom, pos, alt, ref.

titima15 (approved):

titima15:
@ClayBirkett Updating pos, ref, alt works, but updating chrom doesn't work.

ClayBirkett:
@titima15 I removed the updating chrom labeling. If the chrom changes for a marker it will now send a warning. I never intended updating chrom and it would be difficult to implement because of the database structure.

titima15:
The notification: "Markers with mismatched pos, name, ref, or alt compared to the previous...........". It should include only pos, ref or alt. The marker name should not be included here.

ClayBirkett:
I modified the notification message to clarify that the matching to previous marker is done by name. If the chrom, ref, alt, or pos has changed then a warning message will indicate that change.

titima15 (approved):

PR #6176, opened 2026-06-25
Topic/apply plot name template

### LINKED ISSUE #5983: Implement plot name template as trial projectprop
As requested by Mercedes

It would be nice if, like seedlots, plot name templates could be saved to a trial. Users can specify which plot properties are part of the template, including things we may not have in default plot names (rep, block, etc). It is also helpful as a way to automatically regenerate plot names when doing accession swaps.

ryan-preble:
Description 
- Adds autogenerated plot name templates to trial design workflow
- Adds check that autogenerated name templates should produce unique stock names
- tweaks plot name templates to allow text attributes
- adds a preview of template in template creation dialog
- Uses stored templates during accession swaps if there is one

Note: need to decide what to do about multi trial uploads. My thoughts are to add an optional plot_name_template column that accepts the template name. Allows breeding program level control inside the file upload

Closes #5983 #6232 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller:
add optional column with template name

titima15:
@ryan-preble Could you please add autogenerated plot name option (may be as a sub-section) on the first page of Design New Trial dialog and on the spreadsheet info for trial upload?

titima15:
Also, it might be useful to have a link for adding autogenerated name metadata on the Manage Trials page.

ryan-preble:
> @ryan-preble Could you please add autogenerated plot name option (may be as a sub-section) on the first page of Design New Trial dialog and on the spreadsheet info for trial upload?

It is already on the custom plot name section of the trial design workflow. Right now it is on the multi-trial upload only, but I can add it as a conditional option on the single trial upload

lib/SGN/Controller/AJAX/BreedingProgram.pm
```diff
@@ -851,30 +851,7 @@ sub get_autogenerated_name_metadata :Chained('ajax_breeding_program') PathPart('
...
-            my $last_serial_number = $name_metadata->{$format_name}->{'last_serial_number'};
-            my $added_by = $name_metadata->{$format_name}->{'added_by'};
-            my $created_date = $name_metadata->{$format_name}->{'date'};
-
-            my $person = CXGN::People::Person->new($dbh, $added_by);
-            my $person_name = $person->get_first_name()." ".$person->get_last_name();
-
-            push @autogenerated_name_metadata, [$name_type, $format_name, $description, $attributes_string, $last_serial_number, $person_name, $created_date];
+            push @autogenerated_name_metadata, { $format_name => $name_metadata->{$format_name} };
         }
```
titima15, line 859: We usually store who adds data as user_id in the database, then convert to full name for displaying on any page. You can't simplify the code like this, because the table shows user_id, instead of full name (for added by person).

ryan-preble: Thanks for catching that - I kept it as a hash but added another key, added_by_username, and then used that to make a URL on the table display

titima15 (approved):

lukasmueller (approved):
:-)

PR #6190, opened 2026-07-08
Bot CAPTCHA

dwaring87:
Description 

This adds an optional bot captcha to breedbase to try to reduce bots and crawlers from accessing the site.

It requires a running CAPTCHA server (https://trycap.dev/) - there's a docker-compose file included in this PR that can be used to run the server.  Once the server is running, you'll need to create a new key for each instance of breedbase you want to add the captcha too.  Add the URL to the CAPTCHA server and the generated site id/key and secret to the `sgn_local.conf` along with a randomly generated string to enable the captcha:

```
<captcha>
    server             https://cap.example.com
    client_id          xxxx
    client_secret      yyyyyyy
    signing_key        a_randomly_generated_long_string
</captcha>
```

The way this works:

- Every request to the breedbase server that returns an HTML response will check if there is a valid token stored in the user's cookies
- If the token does not exist or is not valid, the request is redirected to the `/captcha` page
- The captcha page will automatically start the verification procedure, including a [proof-of-work stage](https://trycap.dev/guide/effectiveness.html) and an optional [instrumentation stage](https://trycap.dev/guide/instrumentation.html) that are supposed to be computationally expensive enough to deter bots and test the browser environment to check if the script is running in a real browser.  These verification steps don't require any user intervention (no deciphering squiggly letters or finding bicycles).
- The captcha page will generate a response and the breedbase server will check with the CAPTCHA server to verify if the response is correct
- If the response is correct, the breedbase server will sign the response (with the configured signing key) and save the signed response as a token in the user's cookies (`captcha-token`).
- The user will be redirected back to the original page they requested
- Every subsequent request for an HTML page will check if the token is there and the response is correctly signed.
- The verification step will only occur once (as long as the browser's cookies have not been cleared).

To see the verification in action, try going to any of the T3 sites (https://wheat.triticeaetoolbox.org). 

[Screencast from 2026-07-08 14-49-56.webm](https://github.com/user-attachments/assets/5136fcad-b1f5-423e-bd81-ae037537ed54)

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/conf/captcha.docker-compose.yml
+++ b/conf/captcha.docker-compose.yml
@@ -1,0 +1,29 @@
+services:
+  cap:
+    depends_on:
+      valkey:
+        condition: service_healthy
+    image: tiago2/cap:3.1.5
+    container_name: cap
+    ports:
+      - "3000:3000"
+    environment:
+      ADMIN_KEY: change_admin_password
+      REDIS_URL: redis://valkey:6379
+    restart: unless-stopped
+
+  valkey:
+    image: valkey/valkey:9-alpine
+    container_name: cap-valkey
+    volumes:
+      - valkey-data:/data
+    command: valkey-server --save 60 1 --loglevel warning --maxmemory-policy noeviction
+    healthcheck:
+      test: ["CMD", "valkey-cli", "ping"]
+      interval: 5s
+      timeout: 3s
+      retries: 5
+    restart: unless-stopped
+
+volumes:
+  valkey-data:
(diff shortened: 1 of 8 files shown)


ryan-preble (approved):

lukasmueller (approved):
Works after exposing the captcha server correctly on the www.
:-)

PR #6191, opened 2026-07-10
Topic/job fixes

ryan-preble:
Description 
---------------------------------------------------------
Removes the job_finish_log since it was too problematic. Replaces the finish timestamp logic with a script that writes timestamps directly to the database. 
- Important change: background jobs submitted through CXGN::Job must now have db connection params and $basepath to run the finish timestamp script.
- Needs thorough SolGS testing

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/bin/record_finish_timestamp.pl
+++ b/bin/record_finish_timestamp.pl
@@ -1,0 +1,95 @@
+#!/usr/bin/perl
+
+=head1
+
+record_finish_timestamp.pl
+
+=head1 SYNOPSIS
+
+record_finish_timestamp.pl -H [dbhost] -D [dbname] -P [dbpass] -U [dbuser] -j [jobid]
+
+=head1 COMMAND-LINE OPTIONS
+ARGUMENTS
+ -H host name (required) Ex: "breedbase_db"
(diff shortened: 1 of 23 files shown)


isaak (approved):
👍

lukasmueller (approved):
:-)

lukasmueller:
Question: Why are the dbhost, dbname, etc parameters to the submit function, instead of being parameters to the constructor?

ryan-preble:
> Question: Why are the dbhost, dbname, etc parameters to the submit function, instead of being parameters to the constructor?

Because CXGN::Job objects can record completions even if no background script was ever submitted to slurm. For those Jobs, db connection parameters would be unnecessary. 

An alternative I could implement, if you like, is to make them optional parameters to the constructor and have `submit()` die when it does not have the db connection parameters it needs.

ryan-preble:
Revisit the BLAST code to make sure CXGN::Job works there

lukasmueller (approved):
:-)

Issue #6206, opened 2026-07-15
ImageAnalysis Instance: BrAPI-synced images from Field Book are not linked to trial


hkmanching:
In ImageAnalysis, there have been many images synced via BrAPI to the Trial Twist_01. 

The images show up under:
Analyze -> Image Analysis
Plot Page
Search -> Images
Trait Page

However, they are not linked to a trial in the Image Analysis, Images, in the Search Wizard, and do not link to trials in the trait page.

If you look at the actual image page, there is a trial named for an image. 

We need to be able to see the images in the associated trial page and trace the linkage from the Image Analysis, Images, and Search Wizard pages. 

Links:
https://image-analysis.breedbase.org/tools/image_analysis  (Image Analysis Page)
https://image-analysis.breedbase.org/stock/42761/view (Plot page showing image)
https://image-analysis.breedbase.org/image/view/2436 (Image Page showing trial)
https://image-analysis.breedbase.org/breeders/trial/172?format= (Trial page missing images)
https://image-analysis.breedbase.org/search/images (Search Images page showing all images)
https://image-analysis.breedbase.org/cvterm/79295/view (Trait Page)

Screenshot of Search Wizard

<img width="1453" height="916" alt="Image" src="https://github.com/user-attachments/assets/29b45947-447d-414a-a089-702a03169633" />

PR #6215, opened 2026-07-17
Fix multiple image search and include associated field trial in search results

bm743:
Description 
Fixes image search requests with multiple image names or ids as parameters, which fixes the images section on trial details pages. Also includes linked field trials associated to the image in image search results table. 
Adds cvterm trial_associated_image, and images associated with accessions uploaded on a trial page gets associated with that trial, fixing the image section for images on accessions for trial pages

#6206

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


ryan-preble (changes requested):
I might be missing something but I was not able to get this to work. I created a trial of a single accession ID 38846 in the test fixture. I used an image file with EXIF data that you provided to me for an earlier PR test and uploaded it via the trial upload images section. The image was stored in association with the accession but not the trial. 

Image

bm743:
@ryan-preble Thanks for catching this, this was happening because accessions could be associated with multiple trials, so image search for accession based on trial did not work. The new commit should fix this

ryan-preble:
Expected behavior: retrieving images for a trial should get plot images, subplot images, etc. Retrieving images for an accession should get accession-specific images, but also images for plots/subplots/plants this accession was used in. On accession detail page, related stock images should be trial-level images like plots/subplots

ryan-preble:
Tried it again by uploading an image with a plot_id in the EXIF data - the image shows up under the trial images section and related stock images on the accession detail page, but still not on the image search page when searching using the trial to which the plot belongs.

bm743:
Image search by trial should work now

lukasmueller (approved):
This seems to work as advertised!
 :-)

ryan-preble (approved):

PR #6216, opened 2026-07-20
enforce use of dataset genotype protocol

isaak:
Description 
-- ensure the genotyping protocol of a dataset is used in all analyses using genotype data. If a dataset does not have a genotyping protocol property, then use the default or whatever the user selects from the analysis interface. 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


ryan-preble (approved):

lukasmueller (approved):
:-)

Issue #6222, opened 2026-07-23
Failing Drone Imagery Tests


dwaring87:
The `opencv-python` module has recently had a major version change from 4 to 5 that will likely need some manual fixes to the DroneImageScripts ([migration guide](https://github.com/opencv/opencv/wiki/OpenCV-4-to-5-migration)).

To get the tests to pass, you can force reinstall the last version 4 release ([4.13.0.92](https://github.com/opencv/opencv-python/releases/tag/92)):

```
pip install --force-reinstall -v "opencv-python==4.13.0.92"
```

The Dockerfile for building the base image will need to be updated to specify this version until the DroneImageScripts are updated to work with version 5.

PR #6223, opened 2026-07-27
Tweak drone image scripts

ryan-preble:
Description 
Adds an error readout if a script in DroneImageScripts fails. Also restructures some drone image code in SGN that used string-evals, which the linter doesn't like. The actual migration to OpenCV 5 happens in the DroneImageScripts repo. https://github.com/solgenomics/DroneImageScripts/pull/1 

Relevant to #6222 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


bm743 (approved):

lukasmueller:
Could you add a unit test for LegendrePolynomials.pm ?

lukasmueller:
Review by Claude:
Review: [solgenomics/sgn#6223](https://github.com/solgenomics/sgn/pull/6223) — "Tweak drone image scripts"

Stats: +184/−357 across 4 files. Author: ryan-preble. Base: master ← topic/update_drone_imagery_scripts.

Overview

Two independent changes bundled together:

Error handling for RemoveBackground.py / RemoveBackgroundPercentage.py calls in DroneImagery.pm — checks system() exit status and output-file existence, dies with a descriptive message on failure. This is prep work for the OpenCV5 migration happening in the companion [DroneImageScripts#1](https://github.com/solgenomics/DroneImageScripts/pull/1) PR.
Refactor of string-eval code into real subs: a new CXGN::DroneImagery::LegendrePolynomials module replaces inline eval "1/2*(5*$time**3...)"-style formula strings, and a new %env_sim_subs dispatch table + _simulate_environment_value() in DroneImageryAnalyticsRunSimulation.pm replaces eval $env_sim_exec->{$env_simulation}. Both were flagged by the linter for using string eval.

I traced every call site (5 calls to perform_drone_imagery_analytics, 4 calls to _simulate_environment_value) and confirmed the two dropped parameters ($legendre_coeff_exec_array, $env_sim_exec) are removed consistently from the signature and all 5 callers — arg counts match exactly (82 params/args each). Each _simulate_environment_value call site passes only the args its hardcoded $env_simulation value needs (linear_gradient, random_1d_normal_gradient, random_2d_normal_gradient, random — each set right before its use), so the sparse arg hashes at each call site are correct, not a latent bug.

Notable finding: undocumented math fix

The old P5 Legendre term (duplicated in both DroneImagery.pm and DroneImageryAnalytics.pm) was:

'1/16*(63*$time**5 - 70*$time**2 + 15*$time)*$b'

The correct Legendre P5(x) = (1/8)(63x⁵ − 70x³ + 15x). The old code had two bugs: wrong leading coefficient (1/16 vs 1/8) and wrong exponent ($time**2 vs $time**3). The new consolidated LegendrePolynomials.pm (line 658) fixes both:

sub { my ($b, $time) = @_; return 1/2*(5*$time**3 - 3*$time)*$b; },   # P3, unchanged
sub { my ($b, $time) = @_; return 1/8*(35*$time**4 - 30*$time**2 + 3)*$b; },  # P4, unchanged
sub { my ($b, $time) = @_; return 1/8*(63*$time**5 - 70*$time**3 + 15*$time)*$b; },  # P5, FIXED

This silently changes the numerical output of any random-regression analysis using 5th-order (or higher, since order affects everything downstream) Legendre polynomial fits — it's a real correctness improvement, but it's not mentioned anywhere in the PR description, and this is a scientific data platform where reproducibility of past analysis results matters. Worth calling out explicitly in the PR description/commit message so reviewers and downstream users know results for order-5 fits will change.

Other observations
foreach $a (...) → foreach my $a (...) (DroneImageryAnalytics.pm, ~6 spots): good catch — using the un-lexicalized global $a in a loop risks colliding with sort's special $a/$b variables elsewhere in the package. Legitimate bug fix, bundled quietly into an otherwise "refactor-only" PR.
Redundant eval removed (DroneImageryAnalyticsRunSimulation.pm line ~277): eval $sim_data_check_2_times{...} on an already-numeric value is simplified to a plain read. Correct simplification, no behavior change (evaling a plain number was already a no-op).
Error-handling coverage is narrow: the file has ~34 other system($cmd) call sites invoking other DroneImageScripts, none of which get the same exit-status/output-file check added here. That may be intentional scoping (only the two scripts touched by the OpenCV5 migration), but if the goal is "catch script failures," it's a partial fix — worth confirming with the author whether the other call sites are follow-up work or out of scope.
No test coverage added for the new LegendrePolynomials.pm module, despite it now containing the corrected math. Given the silent behavior change identified above, a small unit test asserting legendre_polynomial_sum against known values (especially the P5 term) would guard against regressing this fix later. I didn't find an existing unit test file for CXGN::DroneImagery::* beyond a broad ValidateAllDroneImageryFunctions.t fixture test — worth checking if that fixture exercises order-5+ fits at all.
Style/docs: LegendrePolynomials.pm has proper POD (=head1 NAME/USAGE/DESCRIPTION, =head2), consistent with the project's Perl conventions and the PR checklist's "document new Perl using perldoc" item.
Risk assessment

Low risk overall — the refactor is mechanically verified consistent (param counts, dispatch args), and the error-handling addition is a straightforward defensive check. The one thing I'd want the author to explicitly flag before merge is the P5 formula fix, since it changes computed results for existing users rather than being pure refactoring — that's the kind of change that should be called out, not discovered by a reviewer diffing formula strings.

ryan-preble:
> Could you add a unit test for LegendrePolynomials.pm ?

Yes, I will add that to this PR.

lukasmueller (approved):
:-)

PR #6230, opened 2026-07-29
Ordering tool for cassbase

titima15:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


MercedesThieme:
I tested it, and it works like a charm! However, I noticed a few small things:

- What happens if the status is set to “Cancelled”? It doesn’t move to the Completed tab, but it also can’t be deleted. I assume it would therefore stay indefinitely in the Open Orders section.
- When I select lines from the catalogue and set the number of plants for some or all lines, then untick a line from the catalogue (rather than using the Remove button), all previously entered plant quantities are cleared. The lines themselves remain in the order. When using the Remove button, the quantities are retained.
- The same thing happens when I add a new line by ticking its checkbox: the order form forgets all previously entered quantities.

MercedesThieme:
It works perfectly fine now!

MercedesThieme (approved):

lukasmueller (approved):
:-)

lukasmueller:
This needs to be covered in the manual. Will create an issue for the manual modifications.

PR #6236, opened 2026-07-31
Better error message for missing accession

ryan-preble:
Description 
Checks for undefined resultset, which happens if accession names do not exist. Catches error and reports it more meaningfully to the user

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lib/SGN/Controller/AJAX/TrialMetadata.pm
```diff
@@ -3166,17 +3166,24 @@ sub replace_plot_accession : Chained('trial') PathPart('replace_plot_accessions'
         };
     }
 
-
     my $plot_of_type_id = SGN::Model::Cvterm->get_cvterm_row($schema, 'plot_of', 'stock_relationship')->cvterm_id();
     my $accession_rs = $schema->resultset("Stock::Stock")->search({
         uniquename => $new_accession
     });
+    if (!$accession_rs) {
```
lukasmueller: This might work but I think it would be nicer to be more explicit, such as
```
if ($accession_rs->count() == 0) { ...
```

lukasmueller (approved):

bm743 (approved):

PR #6243, opened 2026-08-06
Fix exif image verification with no trait ID

bm743:
Description 
Fixes verification for exif data that does not have the observation variable id field, falls back to observation variable namae

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/SGN/Controller/AJAX/Image.pm
+++ b/lib/SGN/Controller/AJAX/Image.pm
@@ -386,25 +386,41 @@ sub verify_exif_POST {
                 }
             }
             # Get cvterm_id of recorded trait
+            my $observation_variable = $decoded_json->{observation_variable};
             my $trait_id = $decoded_json->{observation_variable}->{external_db_id};
+            my $trait_name = $observation_variable->{observation_variable_name};
+            print STDERR "trait id test: $trait_id";
             my $cvterm_id;
+            my $trait_lookup_type;
+            my $trait_lookup_value;
 
-            if (!$trait_id) {
-                my $trait_name = $decoded_json->{observation_variable}->{observation_variable_name};
+            if ($trait_id) {
+                $trait_lookup_type  = "id";
+                $trait_lookup_value = $trait_id;
+                $cvterm_id = SGN::Model::Cvterm->find_trait_by_id($schema, $trait_id);
+            }
+            elsif ($trait_name) {
+                $trait_lookup_type  = "name";
+                $trait_lookup_value = $trait_name;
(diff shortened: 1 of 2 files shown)


lukasmueller (approved):
:-)

ryan-preble (commented):
Maybe I don't understand this right - but it seems like in the case `$trait_id` is defined, the code never enters the `if(!$trait_id)` block and `$cvterm_id` is left undefined. Wouldn't that `else` block belong on the outside of the `if` block to directly assign `$cvterm_id` if `$trait_id` is defined as expected?

lib/SGN/Controller/AJAX/Image.pm
```diff
@@ -387,7 +387,18 @@ sub verify_exif_POST {
...
-            my $cvterm_id = SGN::Model::Cvterm->find_trait_by_id($schema, $trait_id);
+            my $cvterm_id;
+
+            if (!$trait_id) {
+                my $trait_name = $decoded_json->{observation_variable}->{observation_variable_name};
+                my $cvterm = $schema->resultset('Cv::Cvterm')->find({ name => $trait_name });
+                if ($cvterm) {
+                    $cvterm_id = $cvterm->cvterm_id();
+                }
+                else {
```
lukasmueller: should be with the if (!$trait_id) { block

ryan-preble (approved):

lib/SGN/Controller/AJAX/Image.pm
```diff
@@ -387,7 +387,18 @@ sub verify_exif_POST {
...
             # Get cvterm_id of recorded trait
             my $trait_id = $decoded_json->{observation_variable}->{external_db_id};
-            my $cvterm_id = SGN::Model::Cvterm->find_trait_by_id($schema, $trait_id);
+            my $cvterm_id;
+
+            if (!$trait_id) {
+                my $trait_name = $decoded_json->{observation_variable}->{observation_variable_name};
+                my $cvterm = $schema->resultset('Cv::Cvterm')->find({ name => $trait_name });
+                if ($cvterm) {
+                    $cvterm_id = $cvterm->cvterm_id();
```
ryan-preble, line 407:
```
                    $cvterm_id = $cvterm->cvterm_id();
```
What happens if $cvterm is undefined? What happens if the $trait_name was a trait synonym?

lukasmueller:
Implement error or informational messages in the case the trait is not in the db or if there is no trait info at all

ryan-preble (approved):

PR #6252, opened 2026-08-10
Add fine-grained cvterm edit config key

### LINKED ISSUE #6245: specify multiple ontologies that can be edited
Currently, the system allows to edit only the first trait_ontology as returned from get_root_names function.
Many systems have multiple trait ontologies, not all of which should be user editable. 
 * define the editable ontologies as a list of db_names with the allow_trait_edits variable
 * display the editable ontologies on the ontology browser on the edit trait page
 * allow to add terms only to the editable ontologies
 * if allow_trait_edits is 1 (a currently legal value), then the default trait ontology should be displayed

ryan-preble:
Description 
Changes the behavior of the config key `allow_trait_edits` to be a list of cvterm DB names for ontologies that can be edited.
Other changes:
- `allow_treatment_edits` is deprecated
- cvterm autocomplete now accepts multiple db names for autocompletion

Closes #6245 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/js/source/legacy/CXGN/BreedersToolbox/AddTrial.js
+++ b/js/source/legacy/CXGN/BreedersToolbox/AddTrial.js
@@ -524,7 +524,7 @@ jQuery(document).ready(function ($) {
     function apply_treatment_autocomplete() {
         jQuery('.treatment-name-input-box').each( function() {
             jQuery(this).autocomplete({
-                source : '/ajax/cvterm/autocompleteslim' + "?db_name=_TREATMENT",
+                source : '/ajax/cvterm/autocompleteslim' + "?db_names=EXPERIMENT_TREATMENT,COMP_EXP_TREATMENT",
                 appendTo : '#add_project_dialog'
             });
         });
--- a/lib/CXGN/Onto.pm
+++ b/lib/CXGN/Onto.pm
@@ -118,7 +118,7 @@ sub get_root_nodes {
                     JOIN dbxref USING(dbxref_id)
                     JOIN db USING(db_id)
                     LEFT JOIN cvterm_relationship ON(cvterm.cvterm_id=cvterm_relationship.subject_id)
-                    WHERE cvterm_relationship.subject_id IS NULL AND cvterm.is_obsolete= 0 AND cvterm.is_relationshiptype = 0";
+                    WHERE cvterm_relationship.subject_id IS NULL AND cvterm.is_obsolete= 0 AND cvterm.is_relationshiptype = 0 AND db.name <> '' AND db.name <> 'null'";
 
       my $h = $self->schema->storage->dbh->prepare($query);
       $h->execute($cv_type);
--- a/lib/CXGN/Trait.pm
+++ b/lib/CXGN/Trait.pm
@@ -796,6 +796,7 @@ sub delete_existing_synonyms {
 sub interactive_store {
 	my $self = shift;
     my $parent_terms_json = shift;
+	my $db_name = shift || die "No trait ontology supplied to store the new term in.\n"; # db name of the ontology the new term belongs to, e.g. CO_334
 
     my $schema = $self->bcs_schema();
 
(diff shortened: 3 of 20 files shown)


lukasmueller:
Claude Review:

Review: [PR #6252 — "Add fine-grained cvterm edit config key"](https://github.com/solgenomics/sgn/pull/6252)
Overview

Replaces the boolean allow_trait_edits/allow_treatment_edits config flags with a single allow_trait_edits key that's either 1 (shorthand: edit only the first/default trait ontology) or a comma-separated list of editable ontology DB names. allow_treatment_edits is folded in — treatment editing is now gated by including EXPERIMENT_TREATMENT in that list. Also generalizes the cvterm-autocomplete endpoints to accept multiple db_names instead of a single db_name. This is access-control code, so I reviewed it closely rather than just skimming the diff.

Critical bug — will break trait creation entirely

lib/SGN/Controller/AJAX/Trait.pm:39:

my $parent_dbs = $c->req->pram('parent_dbs') ? $c->req->pram('parent_dbs') : undef;

pram is not a method on Catalyst::Request (I checked — no such method exists anywhere in this codebase, no AUTOLOAD). This will throw Can't locate object method "pram" via package Catalyst::Request on every single call to /ajax/trait/create, i.e., this breaks trait creation completely. This needs to be $c->req->param('parent_dbs') (twice, on the same line). Given there's no test coverage in this PR touching create_trait, this typo wouldn't be caught by CI — it needs to be fixed and manually exercised before merge.

Authorization logic — real concerns

Since this PR is specifically about tightening/generalizing edit permissions, the exact matching logic matters:

create_trait's permission check uses substring regex matching, not exact list membership:
if (! ($editable_ontologies_str == 1 || (grep {$parent_dbs =~ m/($_)/} @editable_ontologies ) ) ) {

This checks whether $parent_dbs contains an editable ontology name anywhere as a substring, not whether $parent_dbs is (or is drawn from) the editable list. Compare this to edit_trait/delete_trait, which correctly use exact equality (grep {$_ eq $trait->db()}). Two problems:

Substring matching means a short configured name (e.g. COMP) can accidentally match an unrelated, non-editable db name that happens to contain it as a substring.
If $parent_dbs can hold multiple comma-separated parent ontologies (which seems likely, mirroring parent_terms), the check only requires one of them to match — it doesn't verify every referenced ontology is in the editable list. A request could smuggle a restricted ontology through alongside one that's genuinely editable.

I'd recommend splitting $parent_dbs the same way @editable_ontologies is split, and checking exact membership for each element, consistent with how edit_trait/delete_trait do it.

edit_trait constructs the CXGN::Trait object, and evaluates the permission check against it, before validating $cvterm_id (lines ~119-139 in the updated file), whereas the old code validated cvterm_id first and only built the object inside the eval{} later. I checked CXGN::Trait's db attribute — it won't crash on an undef cvterm_id (the underlying DBIC search just returns 0 rows, so db() returns ""), so this won't 500. But it does mean a request with a missing cvterm_id is now routed through the permission check using a bogus empty-string db name before ever hitting the "Cvterm ID missing" validation — fragile ordering for an authorization check, and if allow_trait_edits ever contains a leading/stray comma (e.g. a config typo like ,COMP,CO_334 → splits to ("", "COMP", "CO_334")), a request with no cvterm_id would pass the permission check via "" eq "". Low-likelihood but worth tightening — validate cvterm_id before doing anything permission-related with it.
Default config now silently widens treatment-edit access

sgn.conf changes from:

allow_trait_edits 1
allow_treatment_edits 0

to:

allow_trait_edits EXPERIMENT_TREATMENT,COMP_EXP_TREATMENT,COMP,CO_334

Treatment editing was explicitly off by default before; the new default list includes EXPERIMENT_TREATMENT, turning it on by default for anyone using this reference config. That may well be intentional for this deployment, but it's a permission-boundary change that isn't mentioned in the PR description — worth calling out explicitly so it's a deliberate decision, not an oversight from the config restructuring.

Other observations
db_name= → db_names= migration to autocompleteslim_GET also changes matching from ILIKE '%...%' (substring) to exact equality via regexp_split_to_array+ANY(...). The one call site relying on the old substring behavior (AddTrial.js's db_name=_TREATMENT matching both EXPERIMENT_TREATMENT/COMP_EXP_TREATMENT via substring) was correctly updated to pass the explicit list — good, but this means any other unmigrated caller still passing the old db_name= param would now silently get zero results instead of an error (the controller no longer reads db_name at all). I couldn't exhaustively grep the whole repo for stray call sites from here (no authenticated GitHub code search available), so I'd ask the author to confirm they audited all callers of /ajax/cvterm/autocomplete and /ajax/cvterm/autocompleteslim.
lib/SGN/Controller/Trait.pm: nice incidental fix — the trait-design-page sub was literally named treatment_design_page (an old copy-paste artifact from Treatment.pm), now correctly renamed to trait_design_page.
lib/SGN/Controller/Treatment.pm's treatment_design_page hardcodes db_name => "EXPERIMENT_TREATMENT" only — COMP_EXP_TREATMENT (which AddTrial.js and the default config both treat as a valid treatment ontology) is never passed to the treatment designer template, so it won't be browsable/editable there even when configured as editable. Possibly intentional, but worth confirming with the author.
lib/CXGN/Onto.pm's added AND db.name <> '' AND db.name <> 'null' filter and the trait_designer.mas/treatment_designer.mas swap from /ontology/embedded_browser.mas (single root) to /ontology/browser.mas (multi-root via space-separated root_nodes) are reasonable but I can't verify the root_nodes contract without reading that shared component — worth a manual browser check that the trait designer renders all configured ontology roots correctly.
Test coverage

No test files are included. Given this PR touches authorization logic for trait/treatment creation, edit, and delete, and the pram typo alone shows the create path wasn't exercised, I'd want at least a basic AJAX test hitting /ajax/trait/create, /ajax/trait/edit, /ajax/trait/delete, and /ajax/treatment/create under both the 1 shorthand and an explicit ontology list before merging.

Summary

The design direction (list-based, per-ontology edit permissions) is reasonable and the config migration is mostly consistent, but there's a hard-blocking typo bug (pram) that breaks trait creation outright, plus a real weakening in create_trait's permission check (substring match instead of exact membership) that undermines the fine-grained access control this PR is meant to deliver. I'd fix both before merging.

chris263 (approved):

lukasmueller (approved):
:-)

PR #6255, opened 2026-08-12
Fix additional files table not updating after file delete

### LINKED ISSUE #6221: Representation of additional files is buggy, when deleting a file
observed on CASSbase/Traial Detail Page/Upload data files/Addtional files.

Deleting one file, seems to visually remove all the files. When reloading the page, the files re-appear. 

before deletion:
Image

after deletion: 
Image

after reloading the page: 

Image

ryan-preble:
Description 
Fixes an issue where additional file upload tables were not updating properly after removing a file

Fixes #6221 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/mason/breeders_toolbox/trial/upload_additional_file.mas
+++ b/mason/breeders_toolbox/trial/upload_additional_file.mas
@@ -127,57 +127,48 @@ jQuery(document).ready(function(){
             }
         }
     });
+});		
 
-
- 
-
-
-
-    function display_trial_additional_upload_files (){
-        jQuery.ajax ({
-            url : '/ajax/breeders/trial/'+<% $trial_id %>+'/get_uploaded_additional_file',
-            beforeSend : function(){
-                jQuery('#trial_additional_uploaded_files_div').html("[LOADING...]");
-            },
-            success: function(response){
-                //console.log(response);
-                var html = "<table class='table table-hover table-condensed table-bordered' id='trial_additional_uploaded_files_table'><thead><tr><th>Filename</th><th>Date Uploaded</th><th>Uploaded By</th><th>Options</th></tr></thead><tbody>";
-                for (i=0; i<response.files.length; i++) {
-                    html = html + '<tr><td>'+response.files[i][4]+'</td><td>'+response.files[i][1]+'</td><td><a href="/solpeople/profile/'+response.files[i][2]+'">'+response.files[i][3]+'</a></td><td><a href="/breeders/phenotyping/download/'+response.files[i][0]+'">Download</a> | <a href="javascript:obsolete_additional_file('+<% $trial_id %>+', '+response.files[i][0]+')">Remove</a></td></tr>';
-                }
-                html = html + "</tbody></table>";
-                jQuery('#trial_additional_uploaded_files_div').html(html);
-                jQuery('#trial_additional_uploaded_files_table').DataTable();
-            },
-            error: function(response){
-                alert("Error retrieving additional uploaded files.");
+function display_trial_additional_upload_files(){
+    jQuery.ajax ({
+        url : '/ajax/breeders/trial/'+<% $trial_id %>+'/get_uploaded_additional_file',
+        beforeSend : function(){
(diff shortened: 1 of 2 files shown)


lukasmueller:
Claude review:
Review: [PR #6255 — "Fix additional files table not updating after file delete"](https://github.com/solgenomics/sgn/pull/6255)
Overview

Fixes [#6221](https://github.com/solgenomics/sgn/issues/6221): after deleting ("obsoleting") an uploaded additional file on a trial or stock/accession page, the files table didn't reflect the change. Two Mason templates changed: upload_additional_file.mas (trial) and upload_additional_accessions_file.mas (stock/accessions), both pure <script>-block JS, no Perl/Mason logic touched.

Root cause & fix — verified correct

I pulled both the pre-PR and post-PR versions of upload_additional_file.mas to understand what the whitespace churn was actually doing, since at first glance it looks like pure reformatting. It isn't — it's load-bearing:

Before: obsolete_additional_file(trial_id, file_id) was a globally-scoped function (called via href="javascript:obsolete_additional_file(...)" in the dynamically-built table HTML, which executes in global scope). On success it called jQuery('#...table').DataTable().clear().draw() — this just wipes the DataTable's internal row model in place, without re-fetching from the server, so the table went empty/stale instead of showing the correct remaining files. That's the actual bug.
The natural fix — call display_trial_additional_upload_files() (which re-fetches and rebuilds the table from the server) instead of .clear().draw() — only works if that function is reachable from obsolete_additional_file's scope. But display_trial_additional_upload_files was defined inside the jQuery(document).ready(function(){...}) closure, while obsolete_additional_file was defined outside it (sibling scope, not nested) — so it wasn't callable there.
The fix correctly relocates display_trial_additional_upload_files (and, for consistency, obsolete_additional_file) to top-level/global scope, closing jQuery(document).ready(...) immediately after its initial setup code (line 130 in the new file), so both functions live in the same global scope and can call each other. This is why the diff looks like a big reformat — it needed the brace restructuring to make the actual one-line fix (.clear().draw() → display_trial_additional_upload_files()) valid. Good, well-targeted fix.
Second file — the fix targets unreachable code

For upload_additional_accessions_file.mas, I checked the full file: display_stock_additional_upload_files and obsolete_additional_file are both already defined inside the same jQuery(document).ready(...) closure (siblings within one scope), so no relocation was needed there, and the diff is just the equivalent one-line swap.

However — I searched the entire template (both the JS and the surrounding HTML) and obsolete_additional_file is never actually invoked anywhere in this file. The dynamically-built table HTML (line 140) only renders a "Download" link per row — there's no "Remove"/obsolete link or button in the generated markup, and no other call site exists. So this half of the PR is a correct, harmless fix, but it currently has no observable effect since the delete path it improves isn't reachable from the accessions-file UI at all. Worth a quick sanity check with the author on whether a "Remove" link was intended to ship alongside this (maybe in a follow-up), or whether this half of the change is just precautionary consistency with the trial-file fix.

Minor
for (i=0; ...) in both files declares i without var/let, leaking a global — pre-existing in both files, not introduced by this PR, not worth blocking on.
No tests were added (none existed for this area before either — consistent with this being a client-side-only Mason/jQuery fix). I can't spin up the app from here, so I'd recommend the author (or reviewer) manually click through: upload a file → delete it → confirm the table refreshes to show the remaining files, on the trial page specifically, since that's the reachable path.
Security

No concerns — no new user input handling; trial_id/file_id/stock_id are interpolated the same way as before (pre-existing pattern in this legacy code).

Summary

The trial-file fix (the actual reported bug) is correct and non-obvious — the brace relocation is necessary, not cosmetic, and I verified it against both the before/after source. The accessions-file fix is safe but currently reaches dead code since there's no "Remove" control in that template's UI; flag that with the author rather than blocking on it.

lukasmueller (approved):
:-)

chris263 (approved):

PR #6257, opened 2026-08-12
Fix percent missing calculation on summary statistics

### LINKED ISSUE #6184: "Split by treatment" calculation is incorrect
observed on CASSbase Trial Detail Page/Phenotype Summary Statistics. 

Splitting the table by treatment works now. But the Percentage Missing is calculated incorrectly, as it is based on the whole number of plants, instead of the subset. For example, the ungrouped data shows 0% missing, but after splitting into treatments, the missing data is about 50%, as the other half is part of the other treatment. 
I am also not entirely sure about the coefficient of variance. Assuming CV=SD/Mean, then the number displayed, after splitting by treatment, is not correct. Also, the CV of the ungrouped data is slightly off, but maybe you use a different formula.  

Additionally, the histogram is always shown for the unsplit dataset. However, the little histogram symbol/link in the grouped data table suggests that one might see the split histogram, which is not the case. Either remove this symbol here, or enable a a grouped histogram. 

Image

Image

ryan-preble:
Description 
Fixes the % missing calculation on phenotype summary statistics. For example:
Suppose a trial has 100 plots split evenly between two treatment groups. Suppose each group has 5 accessions of 10 reps each. If 10 plots per group were not measured for a given trait, the table would report 20% missing when grouped by treatments. If not grouped by treatments, 20 plots total would still be unmeasured so still 20% missing. Suppose within each group all 10 plots were of the same accession. Then when grouping by accession the row would read 100% missing for that accession for that trait. 

Note: this does NOT handle the case where treatments were not applied at all to some stocks - I wrote that off as a bad experimental design choice. 

This also disables the histogram link on the summary statistics table when it does not make sense for the histogram button to work (if 100% of data is missing or table is being grouped by accession or treatment)

Fixes #6184 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller:
Claude review:
Review: [PR #6257 — "Fix percent missing calculation on summary statistics"](https://github.com/solgenomics/sgn/pull/6257)
Overview

Fixes the phenotype summary statistics table's "% missing" calculation in phenotype_summary (lib/SGN/Controller/AJAX/TrialMetadata.pm). Addresses [#6184](https://github.com/solgenomics/sgn/issues/6184): when grouping by treatment (and/or accession), the denominator used for % missing was the whole-trial observation-unit count instead of the group's count, and the numerator counted phenotype rows instead of distinct stocks, so repeated measurements could overstate completeness. It also disables the histogram link where clicking it wouldn't make sense, and adds synthetic rows for trait/group combinations with zero measurements that were previously dropped from the table entirely.

Correctness — verified against the full file, not just the diff

I pulled the complete phenotype_summary sub and traced the query logic; the fix is sound:

New $qg group-totals query computes count(DISTINCT plot.stock_id) per group (accession and/or treatment) independent of whether a trait was measured — this is the correct denominator population. Its FROM/JOIN structure (stock → nd_experiment_stock → nd_experiment_project, filtered by project_id/stock_relationship.type_id/plot.type_id/accession.type_id) mirrors the main query's join path, so the group population is consistent with what the main query would return.
count(DISTINCT plot.stock_id) AS n_stocks added to both the numeric ($q1) and non-numeric ($q) trait queries correctly replaces count(phenotype.value) as the "measured" numerator — this genuinely fixes the case where a stock has multiple phenotype records for the same trait (previously could push count above the observation-unit total).
$group_key correctly composes on (accession_id, treatment) so that the same accession under two different treatment values is tracked as two separate groups — this is exactly the scenario described in the issue.
The trailing "missing groups" loop (new lines 759-779) fills in trait×group combinations with zero phenotype rows in either query, so a fully-unmeasured group now shows up as 100% missing instead of being silently dropped from the results — matches the stated intent.
The $percent_missing_for guard (!$group_size || $group_size <= $stocks_measured → "0%") protects against negative percentages if the group total and measured count ever disagree, same defensive shape as the code it replaces.
Bonus fix: display=tissue_samples's ungrouped branch was calling get_subplots() instead of get_tissue_samples() for $total_complete_number — a pre-existing copy-paste bug, correctly fixed here (line 431-432 in the updated file), though it's not mentioned in the PR description.

I don't see a correctness bug in the SQL or Perl — this is a well-reasoned fix, not a guess.

Test coverage — the main gap

The updated test (t/unit_mech/AJAX/BreedersToolbox/TrialMetadata.t:46-47) only re-asserts on group_by_accession=1 with updated expected values (needed because the new "missing groups" loop appends rows at the end of the array, which is why the test switched from checking the last 4 rows to the first 4). However:

No test exercises group_by_treatments=1 at all — the exact scenario in the PR's motivating example (treatment groups) and in issue [#6184](https://github.com/solgenomics/sgn/issues/6184) is untested.
No test isolates the duplicate-measurement fix (a stock with 2+ phenotype values for the same trait) — the unchanged ungrouped-case assertion at line 39 still passes with identical expected values, which only proves the fixture data has no repeated measurements, not that the fix works.
No test explicitly checks the new synthesized 100%-missing row or the histogram-link-disabling logic when percent_missing eq "100%" — the current test's disabled-link rows are disabled because of group_by_accession, not because of the 100%-missing condition, so that branch of the if isn't independently exercised.

Given this PR's entire purpose is fixing a statistics calculation that was silently wrong, I'd want at least one test that actually reproduces the treatment-grouping bug scenario (may require extending the fixture trial with treatment data) before merging — the current test changes verify the refactor didn't break the accession-grouping path, but not that the treatment-grouping bug is fixed.

Minor
The pre-existing stray </a> in the non-numeric-trait loop's disabled histogram span (qq{<span class="glyphicon glyphicon-stats"></span></a>}, line 751) is untouched — the new synthesized-row code correctly omits it (</span> only), so the codebase now has both a correct and an incorrect version of the same disabled-span markup side by side. Not introduced by this PR, but since this code is being touched anyway, worth fixing for consistency.
The PR checklist has no boxes checked at all (not even "Bug fix"), and "The relevant issue has been closed" is unchecked despite "Fixes [#6184](https://github.com/solgenomics/sgn/issues/6184)" in the description — administrative nit, doesn't affect the code.
Security

No concerns — all new SQL uses placeholders for user-influenced values ($trial_id); type IDs are DB-derived cvterm IDs, consistent with the file's existing pattern.

Summary

The calculation fix itself is correct and well-implemented. The blocking concern is test coverage: the PR fixes two specific bugs (treatment-group denominators, duplicate-measurement overcounting) plus adds a new missing-row feature, but the only test change covers the accession-grouping path that was already partially tested before. I'd ask for a treatment-grouping test case before merging.

lukasmueller (approved):
:-)

chris263 (approved):

PR #6259, opened 2026-08-12
update to node.js v24

### LINKED ISSUE #6258: update to Node.js v24
GitHub actions require Node.js v24 for testing. We should also update the breedbase code to use the same Node.js v24

The Webpack v4 we are using for javascript modules is dependent on Node.js v18 so we also need to update to Webpack v5

ClayBirkett:
Updates from node.js v18 to node.js v24
Updates Webpack v4 to Webpack v5, needed for node.js v24

to manually install this update run the following commands inside docker

curl -fsSL https://deb.nodesource.com/setup_24.x | bash -
apt-get update
apt-get install nodejs

cd js
rm node_modules
rm build
npm install
chown -R production node_modules
chgrp -R production node_modules
chmod a+w node_modules/.package-lock.json
npm run build

requires clearing web browser cache

fixes #6258

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -40,6 +40,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HTMLHINT_CONFIG_FILE: .github/linters/.htmlhintrc
           VALIDATE_ALL_CODEBASE: false
+          VALIDATE_SHELL_SHFMT: false
           VALIDATE_ANSIBLE: false
           VALIDATE_CHECKOV: false
           VALIDATE_JSCPD: false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -63,9 +63,9 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Setup Node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
         with:
-          node-version: '18'
+          node-version: '24'
 
       - name: Install JS deps
         working-directory: js
--- a/js/build.webpack.config.js
+++ b/js/build.webpack.config.js
@@ -24,9 +24,11 @@ module.exports = {
         path: path.resolve(__dirname, "build/"),
         publicPath: '/js',
         filename: '[name].min.js',
-        chunkFilename: `chunk.[chunkhash].js`,
-        library: ["jsMod","[name]"],
-        libraryTarget: "umd"
+	chunkFilename: `chunk.[chunkhash].js`,
+        library: {
+	    name: ["jsMod","[name]"],
+            type: "umd"
+	}
     },
     // Set up babel and JSAN processing
     module: {
@@ -59,15 +61,12 @@ module.exports = {
     // Chunks and Minimization settings
     optimization: {
         minimize: true,
-        namedChunks: true,
-        minimizer: [new TerserPlugin({
-            'sourceMap': true,
-            'parallel': 4,
-        })],
-        runtimeChunk: {
+	chunkIds: "deterministic",
+        minimizer: [new TerserPlugin({})],
+	runtimeChunk: {
             name: 'runtime'
         },
-        splitChunks: {
+	splitChunks: {
             cacheGroups: {
                 default: false,
                 shared: {
--- a/js/install_node.sh
+++ b/js/install_node.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 if [[ $(/usr/bin/id -u) -ne 0 ]]; then
-    echo "\nMust run script as root."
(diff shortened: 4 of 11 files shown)


lukasmueller:
Review from Claude:

Review: [PR #6259 — "update to node.js v24"](https://github.com/solgenomics/sgn/pull/6259)
Overview

This PR upgrades the project's JS toolchain from Node 18 → Node 24, which forces a related upgrade of Webpack 4 → 5 (webpack-cli 4→6, terser-webpack-plugin 4→5) for compatibility. It's a focused, 5-file infrastructure change:

.github/workflows/test.yml — setup-node@v4→v5, Node 18→24
js/package.json — bumps webpack/webpack-cli/terser-webpack-plugin, raises engines.node/engines.npm floors
js/build.webpack.config.js — migrates library/libraryTarget to the unified Webpack 5 output.library object, replaces namedChunks with chunkIds: "deterministic", simplifies TerserPlugin options
js/webpack_util/webpack-filemap-plugin.js — rewritten from a prototype-based plugin using the removed compiler.hooks.emit.tapAsync API to an ES6 class using thisCompilation + processAssets
js/webpack_util/jsan-preprocess-loader.js — swaps loader-utils's getOptions() for the built-in this.getOptions()
Code Quality & Correctness
Webpack 5 API migration is correct and non-trivial. The filemap plugin properly accounts for chunk.files becoming a Set in Webpack 5 ([...chunk.files]), switches to compilation.getAsset()/RawSource/emitAsset() instead of directly mutating compilation.assets, and uses PROCESS_ASSETS_STAGE_REPORT — the right stage to read final, already-minified output, matching the old emit-hook timing. Good work.
Good catch, worth calling out explicitly in the PR description: the new code adds if (!lists) continue; before touching legacy_lists[f] (webpack-filemap-plugin.js:189). The old code had no equivalent guard and would have thrown if a chunk emitted a file not present in legacy_lists (e.g., a non-entry asset). This is a real behavioral fix riding along with the refactor — reviewers should know it's intentional, not incidental.
TerserPlugin({}) dropping sourceMap/parallel is correct — devtool: "source-map" is set at the top level of build.webpack.config.js:93, and terser-webpack-plugin v5 derives source-map behavior from devtool automatically; parallel defaults to true in v5, so this isn't a regression.
Style Issues
Mixed tabs/spaces introduced in build.webpack.config.js. Lines 27, 29, 31, 64, 66, and 69 use a leading tab where every surrounding line in the file uses spaces (confirmed with cat -A). This is a stray-whitespace artifact of whatever editor made the edit, not an intentional style choice, and it's inconsistent with the rest of the file. Worth cleaning up before merge, especially if the repo/editor config enforces space indentation.
Suggestions
Remove loader-utils from js/package.json devDependencies. Its only usage in the codebase was in jsan-preprocess-loader.js, which this PR replaces with the built-in this.getOptions(). I checked the rest of js/webpack_util/ and found no other references — the dependency is now dead weight and should be dropped in this PR rather than left for later cleanup.
Risks
Low overall risk — this is a self-contained build-tooling upgrade with no application logic changes. The main risk is the filemap plugin rewrite, since it touches asset-emission internals; the !lists guard and stage choice both look correct on inspection, but this is exactly the kind of change that's cheap to verify and expensive to get subtly wrong, so I'd want to see the CI build actually pass (and ideally confirm mapping.json output is unchanged/expected) rather than relying on code review alone.
No lockfile is committed under js/ and CI uses npm install (not npm ci), so there's no lockfile-drift concern from the dependency bumps.
No other CI workflows (linter.yml, static.yml, test_static.yml) reference Node versions, so nothing else needs updating for consistency.
Test Coverage

No tests were added or modified, which is reasonable for a build-tooling upgrade — the existing CI pipeline (test.yml) exercises npm install against the new toolchain, which is the most meaningful check here. Confirming the CI run is green is more valuable than new unit tests for this kind of change.

Security

Nothing concerning — dependency version bumps are all forward upgrades of mainstream, actively maintained packages (webpack, webpack-cli, terser-webpack-plugin), and removing loader-utils (once cleaned up) reduces the dependency surface slightly.

lukasmueller (commented):
:-)

dwaring87 (approved):
Checked by manually installing node version 24.19.0.  I was able to install the node packages and run the build script.

The changes to the `js/package.json` file do force the system to have node version >= 24, so the [breedbase dockerfile](https://github.com/solgenomics/breedbase_dockerfile/blob/fa1bab6f8ce0cedfa42f79524e89f549cfb73bd3/Dockerfile#L138) will need to be updated to use v24.

lukasmueller (approved):
:-)

PR #6260, opened 2026-08-13
Fix phenotype download

chris263:
Description 

## Problem

  This PR fixes two phenotype-related issues:

  1. Zero-valued observations, were not displayed on the trial details page under “Choose trait and observation level”.
  2. Download Phenotype from Search Wizard caused time out or remain pending indefinitely.

  ## Root causes

  - BrAPI observation serialization used Perl truthiness to validate phenotype values. Since `"0"` is false in Perl, valid zero-valued observations  were excluded.
  - Native searches with `data_level=all` only restricted results through the project join. On large databases, PostgreSQL could build the global phenotype join before applying the trial filter, resulting in a very slow query.

  ## Changes

  - Preserve phenotype values when they are defined and non-empty, including `"0"`.
  - Resolve observation-unit IDs for selected trials before running an all-level Native phenotype search.
  - Apply the observation-unit restriction directly to the main query.
  - Skip rows without a trait while constructing the phenotype matrix.
  - Safely handle undefined start and end dates to avoid related warnings.
  - Add regression coverage for an all-level Native trial search without an explicit plot or plant list.

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


lukasmueller:
Claude review:
Review: [PR #6260 — "Fix phenotype download"](https://github.com/solgenomics/sgn/pull/6260)
Overview

This PR fixes two related phenotype bugs:

Zero-valued observations were being silently dropped from BrAPI serialization because Perl treats the string "0" as false.
Phenotype downloads via Search Wizard timed out on large databases because an unrestricted data_level=all trial search let PostgreSQL build the full phenotype join before applying the trial filter.

Three files changed (+53/−5, 1 commit): Observations.pm, PhenotypeMatrix.pm, Search/Native.pm. No test files are included.

Code Quality & Correctness

Observations.pm (BrAPI serialization filter) — correct fix.

- if ( ($_->{phenotype_value} && $_->{phenotype_value} ne "") || ($_->{phenotype_value} && $_->{phenotype_value} eq '0') )
+ if (defined($_->{phenotype_value}) && $_->{phenotype_value} ne '')

The old condition was actually dead logic: $_->{phenotype_value} eq '0' can never be true when combined with && against the same truthy value, since "0" is false in Perl. So the second clause never fired and "0" was always excluded. The new defined(...) && ne '' check is the right fix and correctly reasoned.

Search/Native.pm (query planner fix) — well-targeted, and I verified it against the rest of the file:

The new pre-query's join (nd_experiment_stock → nd_experiment_project → stock, filtered by project_id IN (trial_ids) and stock.type_id IN (...)) is structurally identical to the main query's existing join path (stock as observationunit → nd_experiment_stock → nd_experiment_project → project, filtered by project.project_id in (trial_sql) at line 582) — same semantics, just computed earlier so Postgres can prune before the expensive phenotype joins. This is a sound optimization, not just a guess.
The 5 type IDs used (plot, plant, analysis_instance, subplot, tissue_sample) exactly match the canonical "all" observation-unit type set already used elsewhere in the file (line ~671-675), so no observation-unit type is silently dropped from results.
$trial_ids are bound via placeholders (?); the type IDs are interpolated directly, but they're cvterm IDs looked up from the DB (not user input) — consistent with the existing pattern throughout this file, not a new injection risk.
The "1 = 0" fallback when the trial has zero matching observation units correctly avoids generating an invalid IN () clause — a small but real correctness detail that other similar list-filters in this file (e.g. lines 548-561) don't guard against, though that's pre-existing and out of scope here.
$has_observationunit_filter only checks plot_list/plant_list/subplot_list, not tissue_sample_list or analysis_instance_list if those exist as separate selectors — worth double-checking there's no such attribute that should also suppress this optimization, though if it doesn't exist this is moot.
Scope limitation: the fix only applies inside if ($self->trial_list && ...). A data_level=all search that filters by e.g. accession_list/program_list/location_list without a trial_list would still hit the original slow path. If the described timeout also reproduces without an explicit trial filter, this PR won't fully resolve it — worth confirming with the reporter that the problematic downloads were always trial-scoped.

PhenotypeMatrix.pm — reasonable but under-explained.

next unless defined($d->{trait_name}) && $d->{trait_name} ne '';

This guards against rows with no trait_name polluting $obsunit_data{...}->{$d->{trait_name}} with an empty-string key. It's a sensible defensive addition, but the PR doesn't explain why such rows now appear (presumably a side effect of the new observation-unit restriction returning obs units with no phenotype data in some path). Since it silently drops rows, I'd want the author to confirm this can't mask a case where legitimate data has a blank trait — low risk, but worth a one-line explanation in the PR description.

Issue: claimed test coverage isn't in the diff

The PR description states "Add regression coverage for an all-level Native trial search without an explicit plot or plant list," but the actual diff (+53/−5, 3 files) contains no test files — only the two .pm source changes plus the matrix guard. The PR checklist also has "Relevant tests have been created and run" unchecked. This is a real discrepancy: either the test commit didn't get pushed, or the description is aspirational. Given this PR is fixing a query-correctness/performance bug, I'd block merge until either the regression test is actually added or the description is corrected — the data_level=all + trial-only branch is exactly the kind of logic that's easy to silently break later without a test pinning its behavior.

Minor
The start_date/end_date caching ($requested_start_date/$requested_end_date) is a good small fix — avoids calling $self->start_date() repeatedly and fixes an uninitialized-value warning risk in the print STDERR when unset.
No EXPLAIN ANALYZE before/after was shared to substantiate the timeout fix — not blocking, but since this targets a specific production timeout, a quick before/after query plan in the PR description would strengthen confidence this actually resolves the reported issue on a representative dataset.
Security

No concerns — all interpolated values are either DB-sourced cvterm/stock IDs or passed through DBI placeholders.

Summary

The core logic in all three files is correct and, for the SQL change, notably well-matched to existing query structure. The main blocker is the mismatch between the PR description's claim of "regression coverage" and the actual diff containing no tests — that should be resolved (either push the missing test commit or correct the description) before merging.

chris263:
@lukasmueller I've fixed all topics identified by claude. It is ready for the next review.

ryan-preble (approved):

lukasmueller:
I ran another Claude review and it came up with these issues, not sure if they are relevant for this pull request. (I actually think that reporting the zero phenotypes might be the behavior that we want):
```
Code review
4 findings
ObservationTables.pm

Confirmed
lib/CXGN/BrAPI/v2/ObservationTables.pm:67.BrAPI table totalCount inflated by empty obs units
BrAPI observation-table totalCount now counts observation units with zero recorded phenotypes, inflating pagination metadata, and this path has no test coverage.
A client calls GET /brapi/v2/observations/table (or /observationunits/table, same bug at ObservationTables.pm:150) with only studyDbIds set. data_level defaults to 'all' and plot_list stays empty (only set from observationUnitDbId), so it hits Native.pm's new all-level prefetch path. That prefetch is precisely what now lets large-trial 'all' queries complete instead of timing out (per the PR's own stated problem), so observation units with no recorded phenotype now reliably come back as rows. PhenotypeMatrix.pm's second loop (unmodified by this PR) unconditionally includes every observation unit in the output matrix regardless of the new 'next unless trait_name' guard, which only skips trait-value columns, not the row itself. total_count = scalar(@data)-1 then counts these phenotype-less rows too, inflating totalCount/totalPages returned to BrAPI API consumers. No test in the diff covers total_count/pagination for this scenario.
lib/CXGN/BrAPI/v2/ObservationTables.pm
:67
Native.pm

Confirmed
lib/CXGN/Phenotypes/Search/Native.pm:355.Duplicates existing get_observation_units_direct() helper
New raw SQL to resolve a trial's observation-unit stock_ids duplicates the existing CXGN::Project::get_observation_units_direct() and get_tissue_samples() helpers.
CXGN::Project::get_observation_units_direct($stock_type) (lib/CXGN/Project.pm:5416-5434) already runs the identical stock/nd_experiment_stock/nd_experiment_project join per trial and stock type, and get_tissue_samples() (Project.pm:5528) covers the one type it doesn't handle by name. The new block hand-rolls the same join instead of composing these existing methods (which are already called a few lines above, in the same per-trial loop, for get_plots()/get_plants()/get_subplots()), leaving two places that must be kept in sync if the observation-unit join logic ever changes.
lib/CXGN/Phenotypes/Search/Native.pm
:355

Plausible
lib/CXGN/Phenotypes/Search/Native.pm:354.New special-cased path diverges from data-level loop
The new 'all'-level observation-unit resolution uses a separate raw-SQL path instead of the existing per-data-level loop/helpers, adding a second divergent mechanism to an already ~640-line function.
The existing per-trial loop (Native.pm:284-339) already resolves plot/plant/subplot ids via CXGN::Trial->get_plots()/get_plants()/get_subplots() for every other data_level. The new 'all' case instead runs hand-built DBI SQL right after that loop rather than extending it, so a future contributor adding a stock type or changing how observation units are resolved for a trial now has to update two differently-shaped code paths instead of one.
lib/CXGN/Phenotypes/Search/Native.pm
:354

Plausible
lib/CXGN/Phenotypes/Search/Native.pm:372.Extra round-trip; large ID list inlined as literal SQL
Re-review: [solgenomics/sgn#6260](https://github.com/solgenomics/sgn/issues/6260) ("Fix phenotype download")
Ran 5 finder agents across 8 angles plus 1 targeted verification pass. This branch already had one round of fixes applied (fixing issues found by claude commit on top of the original Fix phenotype download commit), and the core fix logic held up well — the Perl truthiness fix for "0"-valued phenotypes, the trait-name skip guard in PhenotypeMatrix.pm, and the new observation-unit prefetch in Native.pm are all correctly scoped and don't reintroduce the bugs they're fixing.

4 findings, most severe first:

[correctness, CONFIRMED] ObservationTables.pm:67 — the BrAPI /brapi/v2/observations/table and /observationunits/table endpoints compute totalCount from the phenotype matrix row count. Because the PR's fix now lets large "all"-level trial queries complete (instead of timing out), observation units with zero phenotypes now reliably appear as rows, inflating totalCount/totalPages for BrAPI clients — untested by this PR's own test changes.
[reuse, CONFIRMED] Native.pm:355 — the new raw SQL to resolve a trial's observation-unit stock_ids duplicates CXGN::Project::get_observation_units_direct() / get_tissue_samples(), which already exist and are used a few lines above in the same loop for other data levels.
[altitude, PLAUSIBLE] Native.pm:354 — this special-cases 'all' with a differently-shaped resolution mechanism rather than extending the existing per-data-level loop, adding a second pattern to keep in sync in an already ~640-line function.
[efficiency, PLAUSIBLE] Native.pm:372 — adds a second DB round-trip and re-embeds a potentially large id list as literal SQL text rather than bind params (consistent with existing file convention, so lower severity).
```

chris263:
• Addressed the four findings:
  - Both BrAPI tables exclude unphenotyped rows before pagination, preserving zero values.
  - The all lookup reuses the project helper within the existing trial loop.
  - IDs are deduplicated and bound as a PostgreSQL array.
  - Added regressions and corrected existing response snapshots.

PR #6268, opened 2026-08-25
Allow users to edit transformation ID and notes

titima15:
Description 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change

### DIFF
--- a/lib/CXGN/Transformation/Transformation.pm
+++ b/lib/CXGN/Transformation/Transformation.pm
@@ -41,6 +41,16 @@ has 'transformation_stock_id' => (
     is => 'rw',
 );
 
+has 'transformation_name' => (
+    isa => 'Maybe[Str]',
+    is => 'rw',
+);
+
+has 'transformation_notes' => (
+    isa => 'Maybe[Str]',
+    is => 'rw',
+);
+
 has 'transformants' => (
     isa => 'ArrayRef|Undef',
     is => 'rw',
--- a/lib/CXGN/Transformation/Transformation.pm
+++ b/lib/CXGN/Transformation/Transformation.pm
@@ -590,33 +590,37 @@ sub update_transformation_metadata {
     my $transformation_stock_id = $self->transformation_stock_id();
     my $new_transformation_name = $self->transformation_name();
     my $new_transformation_notes = $self->transformation_notes();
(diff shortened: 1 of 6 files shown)


lib/SGN/Controller/AJAX/Transformation.pm
```diff
@@ -1860,6 +1860,42 @@ sub get_vector_obsoleted_accessions :Path('/ajax/transformation/vector_obsoleted
...
+    }
+
+    my $user_id = $c->user()->get_object()->get_sp_person_id();
+    my $schema = $c->dbic_schema('Bio::Chado::Schema', 'sgn_chado', $user_id);
+
+    my $transformation_obj = CXGN::Transformation::Transformation->new({schema=>$schema, dbh=>$dbh, transformation_stock_id=>$transformation_stock_id, transformation_name=>$new_transformation_name, transformation_notes=>$new_transformation_notes});
+    my $error = $transformation_obj->update_transformation_metadata();
+
+    if ($error) {
+        $c->stash->{rest} = { error => "An error occurred attempting to update the transformation metadata. ($@)" };
```
ryan-preble: While `$@` should work here because nothing overwrites it between here and the eval in `update_transformation_metadata`, it would be a lot safer if the error message was just stored in `$error` and passed along here

ryan-preble (changes requested):

ryan-preble (approved):

lib/CXGN/Transformation/Transformation.pm
```diff
@@ -573,6 +583,48 @@ sub get_transformant_details {
 }
 
 
+sub update_transformation_metadata {
+    my $self = shift;
```
lukasmueller, line 587: Does this need an authorization check?

lukasmueller (commented):
Maybe an authorization check is needed in one of the functions as marked.

lib/CXGN/Transformation/Transformation.pm
titima15: It should be checked in the ajax function

lukasmueller (approved):
:-)

PR #6273, opened 2026-08-25
Manage Locations: add Edit and Remove links to table

### LINKED ISSUE #5992: Managing Locations - make deleting location workflow more obvious to users
You currently have to click on a location on the map in order to delete. Is it possible to make an option to delete from table.

dwaring87:
Description 

This adds Edit and Delete links to the rows in the locations table on the Manage > Locations page

image

Fixes #5992

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


titima15:
@dwaring87 When using the "Edit" link from the table, the original breeding program is not highlighted.

dwaring87:
@titima15 Can you check the latest commit?  I made some changes on how the breeding program options are set as selected and a function to scroll the select box to a selected item so the selection is in view.

titima15:
@dwaring87 It works now.

titima15 (approved):

lukasmueller:
When I edit the location again, it reverts to the old values, unless I reload the page. Maybe there should be an automated page reload triggered, or the data should be refreshed in some way from the database, right after saving it?

dwaring87:
I refactored the JS into a shared CXGN/Locations.js file and added a function to refresh the table and map after an edit.

titima15:
@dwaring87 Now the table is reloaded after updating. However, there is no message indicating that the update is successful. Also, I tested by changing a location name to have the same name as another existing location, the function allows me to do so. Should we allow to have the same location names with different ids?

dwaring87:
I restored the alert saying the location was updated.  I also added name and abbreviation checks to make sure they are unique when an existing location is updated (previously those checks were only done for new locations).

titima15 (approved):

PR #6280, opened 2026-08-28
Field trial linkage remove duplicate libraries

### LINKED ISSUE #6281: remove unneeded javascript libraries
remove unneeded javascript from js/source/legacy/d3
fix trial details, experiment linkage
this should fix weekly emails about vulnerable javascript modules in our repository

javascript loading is already loaded in js/package.json

ClayBirkett:
This will get rid of some of the dependabot alerts listing warning for dangerous code in our repository. The javascript libraries d3-sankey, d3-array, d3-path, d3-shape, are all loaded at build time in package.json.

fixes #6281 remove unneeded javascript libraries

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [x] Bug fix
- [ ] New feature
  - [ ] User-Facing Change
  - [x] Any new _legacy_ JavaScript has been moved from `/js` to `/js/source/legacy`.


ryan-preble (changes requested):
Works well, but a few things:
- lib/CXGN/Stock/Search.pm on line 856 has a `print STDERR Dumper` that absolutely pollutes STDERR and should probably get commented out
- The accession labels in the sankey diagram collide a little bit. It would be good if they could be made smaller or if the thickness of each band could be increased a bit so that the accession labels don't overlap: 

Image

ryan-preble (changes requested):
Good edit changing the sankey bands to one band per trial - but now empty trial types are interpreted incorrectly:

Image

similar problem with the accessions themselves:
image

ryan-preble (approved):

lukasmueller (approved):
:-)

PR #6291, opened 2026-09-09
improve validatio of list for years, projects, Locations

### LINKED ISSUE #2701: Improve list validators
Expected Behavior 
Some list validators are not stringent enough. For example the trial list validator does not distinguish between project types and the year list validator considers any projectprop value a valid year.

For Bugs:
---------

ClayBirkett:
improve query for year and project to match only valid entries
 look for case insensitive matches in locations and report them appropriately

fixes #2701 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


ClayBirkett:
Trials.pm weakens new-trial-name collision checks (medium).
  Several callers use the trials validator inversely — a name in missing means "available to create": TrialGeneric.pm:269, TrialMetadataGeneric.pm:134 (new names), MultipleTrialDesignExcelFormat.pm:509, MultipleTrialDesignGeneric.pm:411. After this change, a new trial name equal to an existing breeding program / trial folder / crossing experiment / analysis now lands in missing and passes the "not already used" check. But project.name carries a DB unique constraint (project_c1), so the subsequent insert fails with a raw DB error instead of the friendly "Trial name(s) … are already used" message. Consider keeping the trials validator strict (project-name existence) and doing the trial-type filtering only where it's semantically wanted, or confirm the callers can tolerate the harder failure

lib/CXGN/List/Validate/Plugin/Trials.pm
```diff
@@ -2,28 +2,60 @@
...
+
+    my %found;
+    if (@unique) {
+        # projectprop / stock_type cvterms that mean "this project is NOT a trial"
+        my @not_trial_type_ids = map {
+            SGN::Model::Cvterm->get_cvterm_row($schema, $_->[0], $_->[1])->cvterm_id()
+        } (
+            [ 'breeding_program',       'project_property' ],
+            [ 'trial_folder',           'project_property' ],
+            [ 'cross',                  'stock_type'       ],  # crossing experiment
```
ryan-preble: I made a list of a trial and a crossing experiment and it still passed validation. These are the projectprops associated with it:

Image

ClayBirkett:
I think the solution is to only use the List validators to check if an entry exist (for example when creating a list) using a specific trial type. The trial creation code should not use this validation, it should compare to a simple list of all trials.

lukasmueller:
There should be a generic stock validator. Would recommend against inlining validation code in tools

PR #6294, opened 2026-09-14
Fix observations brapi call with trialDbId

### LINKED ISSUE #6256: update to brapi observation queries and responses
The user has organized multiple crossing block studies within a trial for a given year. We’d like to be able to quickly query observations for that year without having to loop through each individual study.

I want to do this with a brapi GET query like /observations?trialDbId=XXXX. However, this returns all observations (seemingly), instead of just observations for XXXX. Also, the response does not include a trialDbId field. This isn't stricly required by the brapi schema, but is described in the specification. 

I'm using brapi terms here, assuming that

Brapi study = breedbase trial
Brapi trial = breedbase folder

bm743:
Description 
Fixes #6256 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [ ] New feature
  - [ ] User-Facing Change


ryan-preble (commented):
Code all seems good, but a question about the way Breedbase is meeting BrAPI spec: are studyDbId and trialDbId as breedbase uses them consistent to the way other BrAPI compliant systems use them? What is a BrAPI study vs a BrAPI trial? My gut would tell me a study would be a collection of experiments (trials), so studyDbId would correspond to folder IDs.

ryan-preble:
> Code all seems good, but a question about the way Breedbase is meeting BrAPI spec: are studyDbId and trialDbId as breedbase uses them consistent to the way other BrAPI compliant systems use them? What is a BrAPI study vs a BrAPI trial? My gut would tell me a study would be a collection of experiments (trials), so studyDbId would correspond to folder IDs.

Nevermind - I checked the BrAPI specification and it seems that trials are collections of studies.

ryan-preble (approved):

PR #6296, opened 2026-09-17
fix genotype protocol mixed stock

### LINKED ISSUE #6295: Genotype Protocol improve support for mixed stock
selecting "mixed stocks" is currently labeled as "This is not recommended because it will lead to messy sample metadata"

Problem
1. When selecting "mixed stocks" the current coded incorrectly matches crosses, populations, seedlots, vectors, families. Several of the
2. analysis tools limit genotypes to stocks of accessions and tissue

Fix
When validating genotype uploads restrict is to 5 valid types (accession/tissue_sample/plant/plot/subplot
For analysis tools, modify the genotype search functions to query the 5 valid stock types

ClayBirkett:
fixes for mixed stock in genotype protocols

fixes #6295 

Checklist 
- [ ] Refactoring only
- [ ] Documentation only
- [ ] Fixture update only
- [ ] Bug fix
- [x] New feature
  - [ ] User-Facing Change


ClayBirkett:
Files/functions changed

  lib/CXGN/Genotype/ParseUpload/Plugin/KASP.pm — _validate_with_plugin
  - What it does: checks that the sample names in an uploaded KASP file are valid, using a validator that matches the trial's observation-unit type.
  - What changed: the code used to just add an "s" to the stock type name to pick the validator (e.g. stocks → stockss, which is wrong). Now it explicitly handles the two special cases — tissue_sample_or_accession → tissue_samples_or_accessions, and stocks → stocks (no extra "s") — and only falls back to the old add-an-"s" behaviour for everything else.

  lib/CXGN/Genotype/Search.pm — get_genotype_info and init_genotype_iterator
  - What they do: these two functions fetch genotype records from the database and figure out, for each genotyped sample, which accession (germplasm) it belongs to.
  - What changed:
    - Added new plain read/write attributes on the object (_plot_of_cvterm_id, _plant_of_cvterm_id, _subplot_of_cvterm_id) that simply store the database IDs for the plot_of, plant_of, and subplot_of relationship types, so they only need to be looked up once and can be reused later in the function.
    - The database query that finds a sample's parent accession used to only look for tissue_sample_of relationships. Now it also looks for plot_of, plant_of, and subplot_of, and it only counts a match if the other end of that relationship is actually an accession, cross, or family.
    - The code that fills in the germplasm name/ID used to only run when the sample was a tissue_sample. Now it also runs for plot, plant, and subplot samples.

  lib/CXGN/Genotype/SearchStocks.pm — get_selected_stocks
  - What it does: given a genotyping protocol and a set of accessions, finds the genotyped samples under those accessions.
  - What changed: it used to match samples linked via plot_of, plant_of, or tissue_sample_of. Now subplot_of is included too, so subplot samples are found as well.

  lib/CXGN/List/Validate/Plugin/Stocks.pm — validate
  - What it does: checks that a list of stock names (used for mixed-stock genotype uploads) actually exist and are valid.
  - What changed: it used to check the name against every non-obsolete stock in the database, regardless of type. Now it only checks against stock types that can actually be genotyped — accession, tissue_sample, plant, plot, and subplot — so it can't accidentally match against unrelated things like crosses or seedlots.

  mason/breeders_toolbox/genotyping_trials/upload_genotyping_data_dialogs.mas
  - Just a wording change on the upload screen: the "Mixed Stocks" option no longer warns that mixing stock types is
    discouraged/messy. It now plainly says the sample names can be a mix of accession, plot, subplot, plant, and/or tissue sample names — since that mode is now properly supported.

